PR #16, opened 2019-08-07
Migrate cron jobs to ECS

nsmithuk:


terraform/terraform_environment/ecs_api_cron.tf
```diff
@@ -0,0 +1,82 @@
...
+  assume_role_policy = data.aws_iam_policy_document.cloudwatch_events_assume_policy.json
+}
+
+//---
+
+data "aws_iam_policy_document" "cloudwatch_events_role_policy" {
+  statement {
+    effect    = "Allow"
+    actions   = ["ecs:RunTask"]
+    resources = ["*"]
```
andrewpearce-digital: I think we can limit this to a specific cluster, which would be a good limitation on the policy
`aws_ecs_cluster.online-lpa.arn` should do it.

terraform/terraform_environment/ecs_api_cron.tf
```diff
@@ -0,0 +1,82 @@
...
+  }
+  statement {
+    effect    = "Allow"
+    actions   = ["iam:PassRole"]
+    resources = ["*"]
```
andrewpearce-digital: I think we can limit what roles can be passed here to limit the scope of the policy. That might also make this policy specific to the types of tasks you want to run.

For this task, I think that can be `aws_iam_role.api_task_role.arn`

It might need the execution role arn too.

terraform/terraform_environment/locals.tf
```diff
@@ -37,7 +37,7 @@ locals {
   sirius_api_gateway_endpoint = var.accounts[local.account_name].sirius_api_gateway_endpoint
   sirius_api_gateway_arn      = var.accounts[local.account_name].sirius_api_gateway_arn
 
-  ecs_task_count = 2
+  ecs_task_count = 0
```
andrewpearce-digital: Could you reset this to 2 before merging so preprod and prod are running for migration testing please.

terraform/terraform_environment/ecs_api_cron.tf
```diff
@@ -0,0 +1,157 @@
...
+}
+
+resource "aws_cloudwatch_event_rule" "mid_morning" {
+  name = "${local.environment}-mid-morning-cron"
+  //schedule_expression = "cron(0 10 * * ? *)" // 10am UTC, every day.
+  schedule_expression = "rate(2 minutes)"
+}
+
+//------------------------------------------------
+// Task defernition for Cron
```
cooperaj: Gone a bit Norfolk here

PR #23, opened 2019-08-13
update cpu and memory allocation to match old prod

andrewpearce-digital:
Old production assigned the following instances to each part of the service

Service | EC2 Instance
-- | --
admin | T2.medium
api | T2.medium
front | T2.medium
pdf | T2.small

Instance Type | vCPUs | Memory | ECS CPU | ECS Memory
-- | -- | -- | -- | --
T2.medium | 2 | 4GB | 2048 | 4096
T2.small | 1 | 2GB | 1024 | 2048

Update ECS task definitions to match

### DIFF
--- a/terraform/terraform_environment/ecs_admin.tf
+++ b/terraform/terraform_environment/ecs_admin.tf
@@ -59,8 +59,8 @@ resource "aws_ecs_task_definition" "admin" {
   family                   = "${local.environment}-admin"
   requires_compatibilities = ["FARGATE"]
   network_mode             = "awsvpc"
-  cpu                      = 2048
-  memory                   = 4096
+  cpu                      = 512
+  memory                   = 1024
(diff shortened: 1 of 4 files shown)


terraform/terraform_environment/ecs_admin.tf
```diff
@@ -59,8 +59,8 @@ resource "aws_ecs_task_definition" "admin" {
   family                   = "${local.environment}-admin"
   requires_compatibilities = ["FARGATE"]
   network_mode             = "awsvpc"
-  cpu                      = 512
-  memory                   = 1024
+  cpu                      = 2048
```
nsmithuk: I don't think we need to change the values on `admin`, it'll never have more than 1 user.

andrewpearce-digital: 👍

PR #33, opened 2019-08-17
Async destroy

andrewpearce-digital:
Provides the infrastructure for a queue and lambda function in development only.
Plans for preprod and prod workspace show no resources to create.


andrewpearce-digital:
mark_workspace code successfully sends a workspace name to the queue 
image

PR #40, opened 2019-08-20
Docker Build Steps Update

andrewpearce-digital:
Docker Build now runs from local paths instead of cloned external repositories
Docker Tag now only tags master branch builds with latest
Docker Push, pushes all tags.


andrewpearce-digital:
This run of the pipeline built and pushed branch tags only
note commit hash `d304fcf`
image

ECR doesn't show push tagged as `latest`
image

andrewpearce-digital:
This run of the pipeline built and pushed branch tags and latest
note commit hash 1e6575a
image

ECR shows push tagged as branch tag and latest
image

andrewpearce-digital:
CasperJS Suite run from local and passed
image

PR #44, opened 2019-08-21
Lpa 3345 add unit tests to pipeline

SeemaMenon:


service-pdf/docker/app/Dockerfile
```diff
@@ -10,7 +10,8 @@ FROM php:7-cli
 RUN apt-get -y update \
     && mkdir -p /usr/share/man/man1 \
     && apt-get -y install pdftk \
-    && apt-get -y clean
+    && apt-get -y clean \
+    && pecl install xdebug
```
nsmithuk:
```suggestion
    && pecl install xdebug \ 
    && pecl clear-cache
```

service-front/module/Application/tests/View/Helper/MoneyFormatTest.php
```diff
@@ -18,7 +18,7 @@ public function testInvokeFloor():void
 
     public function testInvokeCeil():void
     {
-        $amount= 50.55;
+        $amount= 50.56;
```
nsmithuk: This seems an odd change. Do we know why it happened?

SeemaMenon: Neil, I made the change as the test was breaking with the error

nsmithuk: @SeemaMenon Sorry, I mean I assume the test was once passing with 50.55. So did something change so that the test needs updating? Or was the test always failing and needing 50.56?

PR #45, opened 2019-08-22
Lpa 3360

andrewpearce-digital:
Read the healthcheck endpoint for the service, make sure it returns a 200, and that the version deployed is expected.

Checking the endpoint from CircleCI requires adding an ingress rule on non-production stacks.

We also need to wait for ECS deployments to settle


andrewpearce-digital:
image

andrewpearce-digital:
Ingress is now managed by terraform with local state
image

Healthcheck tests verify deployed version and status
image

Removing ingress ALWAYS runs
image

andrewpearce-digital:
https://circleci.com/docs/2.0/env-vars/#using-bash_env-to-set-environment-variables

andrewpearce-digital:
Slack messages are now working
image

PR #55, opened 2019-09-09
Removed unnecessary duplicate call in setCertificateProviderSkipped

GemTay:


service-front/module/Application/src/Model/Service/Lpa/Metadata.php
```diff
@@ -27,8 +27,6 @@ public function setReplacementAttorneysConfirmed(Lpa $lpa)
 
     public function setCertificateProviderSkipped(Lpa $lpa)
     {
-        $this->setMetadataByKey($lpa, Lpa::CERTIFICATE_PROVIDER_WAS_SKIPPED);
```
taylorsj1980: Should this definitely be removed? These lines are not the same thing.
Lpa::CERTIFICATE_PROVIDER_WAS_SKIPPED
and
Lpa::CERTIFICATE_PROVIDER_SKIPPED

PR #62, opened 2019-09-17
Lpa 3401 add gds requested cross domain tracking

cooperaj:
Add cross domain tracking to GA on LPA


service-front/assets/js/moj/moj.modules/moj.analytics.js
```diff
@@ -27,9 +27,17 @@
...
-        allowAnchor: true
+        allowAnchor: true,
+
+        //TODO are we tracking this within lpa
+        stripPostcodePII: true,
+        stripDatePII: true
       });
 
+      if (document.domain === 'lastingpowerofattorney.service.justice.gov.uk') {
+        GOVUK.analytics.addLinkedTrackerDomain(gaConfig.govId, 'govuk_shared', ['www.gov.uk', '.payments.service.gov.uk']);
```
gregtyler: Should this just be `payments.service.gov.uk` (without the leading `.`)?

cooperaj: I think that this should cover www.payments.service.gov.uk and payments.service.gov.uk. I could be wrong.

cooperaj: I guess I was going on what was above in that file

```
 var cookieDomain = (document.domain === 'lastingpowerofattorney.service.justice.gov.uk') ? '.lastingpowerofattorney.service.justice.gov.uk' : document.domain;
```

I'm happy to remove but then this one should also be right?

gregtyler: I've just dug into some [documentation](https://developers.google.com/analytics/devguides/collection/analyticsjs/linker#parameters) and yours is quite right: it does a substring match, so the `.` ensures it matches `www.payments.service.gov.uk` but not `some-other-service-payments.service.gov.uk` :+1:

PR #70, opened 2019-09-23
Error page for if user has no account details set in front service

GemTay:


service-admin/src/App/src/ConfigProvider.php
```diff
@@ -63,6 +63,7 @@ public function getDependencies() : array
                 Middleware\Authorization\AuthorizationMiddleware::class => Middleware\Authorization\AuthorizationMiddlewareFactory::class,
                 Middleware\Session\SessionMiddleware::class             => Middleware\Session\SessionMiddlewareFactory::class,
                 Middleware\ViewData\ViewDataMiddleware::class           => Middleware\ViewData\ViewDataMiddlewareFactory::class,
+                Middleware\UserValidation\UserValidationMiddleware::class           => Middleware\UserValidation\UserValidationMiddlewareFactory::class,
```
cooperaj: This should be reindented so all the '=>' line up as they do in the lines above.

service-admin/src/App/src/Middleware/UserValidation/UserValidationMiddleware.php
```diff
@@ -0,0 +1,45 @@
+<?php
+
```
cooperaj: Going forward with PSR-12 (https://www.php-fig.org/psr/psr-12/) you'll only separate lines in the start of a file with one linebreak. We already do this everywhere anyway so one of these should be dropped.

service-admin/src/App/src/Middleware/UserValidation/UserValidationMiddleware.php
```diff
@@ -0,0 +1,45 @@
...
+use Psr\Http\Message\ResponseInterface;
+use Psr\Http\Message\ServerRequestInterface;
+use Psr\Http\Server\MiddlewareInterface;
+use Psr\Http\Server\RequestHandlerInterface;
+use Zend\Diactoros\Response\HtmlResponse;
+use Zend\Expressive\Template\TemplateRendererInterface;
+
+class UserValidationMiddleware implements MiddlewareInterface
+{
+
```
cooperaj: This is an extra line break too.

service-admin/src/App/templates/error/no-user-details-error.phtml
```diff
@@ -0,0 +1,9 @@
+<?php $this->layout('layout::default', ['title' => 'Incomplete account details']) ?>
+
+<h1>Oops!</h1>
+<h2>This is awkward.</h2>
+<p>You need to complete your LPA account details first!</p>
+<p>
+    You must fill in your account details on the LPA online service before accessing the Admin service.
+</p>
+<p> Click <a href="https://www.lastingpowerofattorney.service.gov.uk/login">here</a> to sign into your LPA account and complete your details.</p>
```
cooperaj: A better way to do linking in HTML documents is to avoid the use of clicking "here" and instead explain what the link does. For instance;

"You will need to [sign in using your account] on the LPA frontend and complete your details".

Sadly we don't have a content person anymore to take a look at this but theres lots on the web [that you might want to read](http://www.leafdigital.com/class/lessons/writing/9.html).

PR #73, opened 2019-10-04
Lpa 3415 fix cookie consent properly

cooperaj:
Do proper cookie consent checking.


service-front/public/assets/v2/js/application.js
```diff
@@ -23041,6 +23041,190 @@ var SessionTimeoutDialog = function (options) {
...
+                return decodeURIComponent(cookie.substring(nameEQ.length))
+            }
+        }
+        return null
+    }
+
+    global.GOVUK = GOVUK
+}(window))
+;
 this["lpa"] = this["lpa"] || {};
```
GemTay: What's the purpose of these 2 lines here to do with an LPA?

cooperaj: This is a compiled file and can be ignored for the purposes of PRs

Thats said I believe these two lines are specific to the javascript templating components that are built and show here due to the way diffing works.

PR #82, opened 2019-12-05
Updated packages with security vunerabilities

GemTay:


service-front/package.json
```diff
@@ -20,8 +20,13 @@
     "grunt-text-replace": "^0.4.0"
   },
   "dependencies": {
+    "axios": "^0.19.0",
```
cooperaj: Are these dependencies now? or dev dependencies?

GemTay: I believe they are dependancies now

PR #95, opened 2020-03-12
Lpa 3539

andrewpearce-digital:
## Purpose
As a team
We want to add a job to the pipeline to check results of ECR scan and report findings
So that we can act to prevent known vulnerabilities reaching production

In addition, 

- all Terraform syntax has been updated to meet the spec for version 0.12.
- the path for the environment level Terraform has been changed from `terraform_environment` to `environment`
- all scripts have been put into the `scripts` path 
- Terraform has been upgraded to version 0.12.23

Fixes LPA-3539

## Approach

Use python to retrieve ECR scan results once they are completed, then post messages to slack for each finding.

## Learning

Based on work completed on the opg-use-an-lpa project.

## Checklist


andrewpearce-digital:
I've decided to do one more thing on this script, so it's not ready to review yet.

I'm going to set it up so you provide a file with the images to scan instead of it being in the code.

This will make it easier to use on any repository.

jamesrwarren:
Looking pretty great to me.

terraform/environment/.envrc
```diff
@@ -1,4 +1,4 @@
 export TF_VAR_default_role=operator
 export TF_VAR_management_role=operator
 export TF_CLI_ARGS_init="-backend-config=role_arn=arn:aws:iam::311462405659:role/operator -upgrade=true -reconfigure"
-export TF_WORKSPACE=
+export TF_WORKSPACE=95-LPA3539
```
williamfalconeruk: is this supposed to be set?

PR #96, opened 2020-03-17
LPA-3420 ECS Scaling on Demand for Make a LPA

andrewpearce-digital:
## Purpose
Use CPU and Memory metrics tracking to scale ECS services to meet demand (inferred by the load on services).

Fixes LPA-3420

## Approach

Create an IAM role for ECS Autoscaling
Given the duplication of resources required to set up autoscaling for all ECS services, resources will be defined once using a terraform module, and use that module for each service.
In addition, improve the setup and use of environment parameters based on target account.

## Checklist


andrewpearce-digital:
> TF workspace env needs clearing.

spotted and fixed :D

andrewpearce-digital:
Scaling occurred during testing with cpu and memory usage targets
image

Scaling for front end remained at max of 5 due memory usage

No errors during testing with autoscaling
image

andrewpearce-digital:
Auto Scaling set up for each service that requires scaling (admin does not)
image
image
image

terraform/environment/ecs_front.tf
```diff
@@ -59,8 +60,8 @@ resource "aws_ecs_task_definition" "front" {
   family                   = "${local.environment}-front"
   requires_compatibilities = ["FARGATE"]
   network_mode             = "awsvpc"
-  cpu                      = 2048
-  memory                   = 4096
+  cpu                      = 256
```
williamfalconeruk, line 63: were these values too high previously? - just interested....

andrewpearce-digital: Good question.

I did some analysis on our usage on production and redefined the allocation for all of the services to get the same level of service out of smaller tasks (get rid of under-utilisation).

The utilisation min, max and avg were taken from Cloudwatch across a timeframe of all available data (since we migrated to ECS to now).

image

To stay on the safe side, while I've reduced the CPU and memory, I've kept the minimum task count the same for now.

The result should be no change in performance of the service, auto-scaling to deal with peaks, cheaper ECS services and the scope to scale down a bit more when in low demand.

mattmachell: There's potential to do further performance tweaks around how NGINX and PHP interact, but this is really good in terms of making our default scaling appropriate to the needs of the app. Great work!

PR #98, opened 2020-03-23
Allow admin to check records for a user when searching via admin app

SeemaMenon:
## Purpose
https://opgtransform.atlassian.net/browse/LPA-3571

Fixes LPA-3571

## Approach

Access permission was denying users application list to be searched by service call from admin app. Removed the access permission check to test.

## Learning

_Any tips and tricks, blog posts or tools which helped you. Plus anything notable you've discovered about the Refunds service_

## Checklist


cooperaj:
Does this fix not reduce the security of the API?

SeemaMenon:
> Does this fix not reduce the security of the API?

SSM: I did think about, may be we should have a discussion on how this works and alternate possible. Shall pause on this for now.

PR #124, opened 2020-08-04
Lpa 3738 add seeded data for dev environments.

williamfalconeruk:
## Purpose

Fixes LPA-3738

## Approach

- access cloud 9 and preprod DB
- follow runbook to extract data, and select appropriate records for ticket
- add into existing seeding script

## Learning
- Setting up a cloud9 instance.
- Cloud9 needs public subnets to work.
- Cloud9 Postgres default mismatched our RDS instance, so had to be fixed in our scripts; note we used client not full server. see: https://notes.alessiosignorini.com/post/170166752543/install-postgres-on-aws-cloud9 

## Checklist


docs/runbooks/cloud9/cloud9_init.sh
```diff
@@ -66,16 +67,18 @@ function postgresql() {
   local environment_name=$(echo ${1:?} | awk '{print tolower($0)}')
   export AWS_DEFAULT_REGION=eu-west-1
   export PGHOST=$(aws rds describe-db-instances --db-instance-identifier api-${environment_name} | jq -r .'DBInstances'[0].'Endpoint'.'Address')
```
andrewpearce-digital, line 69: An interesting optimisation possible here. The script is now making the exact same api call and using jq to grab the right part to set with export.

If you make the request once and assign it to a variable like db_instance and then use jq against that variable to export the bits we need.

```
db_instance=$(aws rds describe-db-instances --db-instance-identifier api-${environment_name}
  export PGHOST= $(cat $db_instance | jq -r .'DBInstances'[0].'Endpoint'.'Address')
️```
or something like that

williamfalconeruk: good point - will take a look

williamfalconeruk:
Screenshot 2020-08-11 at 16 34 39

PR #131, opened 2020-09-22
Lpa 3762 fix localstack compatibility issues

williamfalconeruk:
## Purpose
Fixes issues with localstack, which were found during running the latest version. This does not affect live systems, because the changes only affect development environments. This can be merged safely once reviewed.

Fixes LPA-3762

## Approach

- Fix port configs to edge port 4566 for localstack
- Workaround issue with http request not returning a value when executing waitforit on 4566. 
- Fix subsequent issue with SQS queue URN configuration changing

## Learning

There have been a couple of breaking changes in LocalStack, which is used for the queuing and file generation of PDFs.

Since 0.11.6, All traffic now has to go through the AWS edge port (4655), rather than as specified for each service.
See:

- https://github.com/localstack/localstack/pull/2905
- https://github.com/localstack/localstack/issues/2983

Note the legacy ports had been deprecated on previous versions.

since 0.11.3 of localstack, SQS queue creation added a fake account id to the urn used for the queue, to be consistent with the live AWS SQS services. this broke the PDF generation messaging queue.

- https://github.com/localstack/localstack/issues/2612 
- https://github.com/localstack/localstack/issues/2877
- https://github.com/localstack/localstack/pull/2591

rather than pinning to a working version, we fixed the breaking docker-compose configuration. there were also  issues with waitforit on the local-config service, which we could workaround using a tcp check only.

## Checklist

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -20,12 +20,11 @@ services:
       - 8000:8000
 
   localstack:
-    image: localstack/localstack:0.9.1
+    image: localstack/localstack:latest
     ports:
-      - 4569:4569
-      - 4576:4576
+      - 4566:4566
     environment:
-      - SERVICES=s3:4569,sqs:4576
+      - SERVICES=sqs,s3
       - DEFAULT_REGION=eu-west-1
       - HOSTNAME=localstack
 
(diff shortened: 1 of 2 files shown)


charlesmmarshall (approved):
lgtm

Raffers (approved):

PR #157, opened 2020-10-15
LPAL-20 - As a mobile device user I would like the appropriate keyboard to appear for numeric only fields

Raffers:
## Purpose
When using the use an LPA service on a mobile phone or a tablet, each time a user tries to enter a numeric field then the default input should be numbers rather than the qwerty keyboard. 

Fixes LPAL-20

## Approach

Updates the appropriate fields with the required attributes. Creates a custom FromText helper to allow `inputmode` attribute to be specified on elements.

## Learning

https://docs.laminas.dev/laminas-view/helpers/intro/
https://docs.laminas.dev/laminas-form/helper/abstract-helper/ 

## Checklist


johnnolan:
An FYI on this, all <form> tags will need a novalidate attribute on them to prevent FF trying to do in line validation https://www.w3schools.com/tags/att_form_novalidate.asp

Raffers:
> An FYI on this, all
> tags will need a novalidate attribute on them to prevent FF trying to do in line validation https://www.w3schools.com/tags/att_form_novalidate.asp

The current plugin we're using to generate the form elements doesn't seem to support newer attributes, it's not applying min/max, inputmode & novalidate

johnnolan (commented):
> > An FYI on this, all
> > tags will need a novalidate attribute on them to prevent FF trying to do in line validation https://www.w3schools.com/tags/att_form_novalidate.asp
> 
> The current plugin we're using to generate the form elements doesn't seem to support newer attributes, it's not applying min/max, inputmode & novalidate

This may be a no go then, we may need to park this ticket until we have the ability to add the novalidate otherwise it will stop people from using the site rather than helping.

service-front/module/Application/view/application/authenticated/about-you/index.twig
```diff
@@ -75,15 +76,17 @@
...
     min: '0',
     max: '12',
-    pattern: '[0-9]*'
+    pattern: '[0-9]*',
+    inputmode: 'numeric'
  }) ? ''}}
 
 {{ yearOfBirth.setAttributes({
     id: 'dob-date-year',
     type: 'tel',
```
townxelliot, line 85: Should this type be "tel"? I assume "tel" is for telephone numbers, but this is a year field.

townxelliot: I realise you didn't add this, just wondering why it's here.

townxelliot:
> > An FYI on this, all
> > tags will need a novalidate attribute on them to prevent FF trying to do in line validation https://www.w3schools.com/tags/att_form_novalidate.asp
> 
> The current plugin we're using to generate the form elements doesn't seem to support newer attributes, it's not applying min/max, inputmode & novalidate

@Raffers which plugin is it? Is it not possible to define arbitrary attributes on an element?

service-front/module/Application/view/application/authenticated/about-you/index.twig
Raffers: All these fields have now been updated to Text

service-front/module/Application/config/module.config.php
```diff
@@ -110,6 +110,7 @@
             'formRadio'             => 'Application\Form\View\Helper\FormRadio',
             'formCheckbox'          => 'Application\Form\View\Helper\FormMultiCheckbox',
             'finalCheckAccessible'  => 'Application\View\Helper\FinalCheckAccessible',
+            'formtext'              => 'Application\Form\View\Helper\FormText',
```
townxelliot, line 113: It would be good to understand a little bit more about how this overrides the default formtext helper. Could you perhaps add a comment or some notes somewhere which explain why we have done this here? This would help give some context to anyone coming along later (e.g. if Laminas ever fix the issue themselves, we'd want to know where in our code we would need to come back to and change).

townxelliot (changes requested):
Looks good to me, but I'd like some more documentation of the decision making in the code, or at least a link to the issue we're working around. In the latter case, the decision process could be outlined in the bug so we understand what's going on here, e.g. "we did this because Laminas ignores the inputmode attribute; we raised this PR with Laminas to fix it upstream".

service-front/module/Application/config/module.config.php
Raffers: I've left a comment in the config file, linking to the issue. But after looking at the documentation I don't think it's best to start putting base Laminas documentation links into the codebase.

The links that helped me though -
https://docs.laminas.dev/laminas-view/helpers/intro/
https://docs.laminas.dev/laminas-form/helper/abstract-helper/ 

I've dropped them in the PR as well.

townxelliot (approved):

nickdavis2001 (approved):
lgtm

PR #178, opened 2020-11-10
Makefile and build speed improvements

nickdavis2001:
## Purpose
_Improve the reset option to clean up properly, add options to only reset front and only run front unit tests. Make dc-run now runs in parallel. composer uses prestissimo in make dc-run_

## Approach

_See source files, fairly self-explanatory. _

## Checklist

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -97,7 +97,7 @@ services:
       OPG_LPA_COMMON_DYNAMODB_ENDPOINT: http://dynamodb:8000
 
       PHP_OPCACHE_VALIDATE_TIMESTAMPS: 1
-      #ENABLE_XDEBUG: 'true'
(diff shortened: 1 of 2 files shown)


docker-compose.yml
```diff
@@ -323,13 +338,18 @@ services:
     image: composer:1.10.16
     volumes:
       - ./service-pdf:/app
+    command:
```
williamfalconeruk, line 341: never knew you could do this!

williamfalconeruk (approved):

charlesmmarshall (approved):

docker-compose.yml
nickdavis2001: Neither did I till about 40 mins ago ;)

PR #186, opened 2020-11-19
Add Cypress tests into build

nickdavis2001:
## Purpose
_Introduce cypress into the testing mix for LPAL-6 JIRA_

## Approach

_Similar to how done for pa11y, has its own docker container which can be spun up locally or by circle. First test included which does a login_

## Learning

## Checklist


charlesmmarshall (approved):

Raffers (approved):

williamfalconeruk (approved):

cypress/Dockerfile
```diff
@@ -0,0 +1,10 @@
+FROM cypress/included:5.2.0
+
+WORKDIR /root
+
+RUN apt-get install curl
```
williamfalconeruk, line 5: looks like this can be safely removed.

williamfalconeruk: the docker image already has this in - that's useful to know 👍

cypress/Dockerfile
```diff
@@ -0,0 +1,10 @@
...
+
+ENV CYPRESS_VIDEO=false
```
williamfalconeruk, line 7: you can pass this in as a -e option when doing `docker run` if need be. this is optional.

williamfalconeruk (changes requested):
just the one change needed to remove that download step, then it's good to go.

cypress.json
```diff
@@ -0,0 +1 @@
+{}
```
williamfalconeruk, line 1: more of a question, i'm guessing this is needed even if empty? is this where we configure cypress?

.circleci/config.yml
```diff
@@ -719,13 +725,17 @@ orbs:
...
+                docker run -it -e "CYPRESS_baseUrl=https://$FRONT_DOMAIN" cypress:latest # cypress run --spec ~/project/cypress/integration/basic_login.js
+          - run:
+              name: Run pa11y tests
               command: |
                 docker run -dt -e AWS_ACCESS_KEY_ID -e AWS_SECRET_ACCESS_KEY -e AWS_SESSION_TOKEN -e CI -e "BASE_DOMAIN=$FRONT_DOMAIN" --name opgpa11y opgpa11y:latest
                 docker exec opgpa11y sed -i "s/localhost:7002/$FRONT_DOMAIN/" config.json
                 docker exec opgpa11y pa11y-ci -c config.json || true # true prevents this being a failure if pa11y has errors. one day we will want to remove true
           - run:
-              name: Run test
+              name: Run caspar tests
```
Raffers, line 738: casper > caspar

cypress/Dockerfile
nickdavis2001: Indeed it is. Good spot :)

williamfalconeruk: yep - circleci logs mentioned no update needed. happy to approve now :)

williamfalconeruk (approved):

cypress/Dockerfile
nickdavis2001: Indeed.  This came across with Vega's dockerfile. I guess its a sensible default as the only time we want the video is when running locally.

cypress.json
nickdavis2001: I think its still needed even inf empty, but in future will put the localhost address in there actually so that we can run locally outside docker

PR #195, opened 2020-11-27
LPAL-13 - Cookie settings send user to home page

Raffers:
## Purpose
LPA: When a user clicks on the cookie settings link in the banner or at the page footer, once they have saved their cookies preferences/changes they do not return to the page they were on prior. Instead, they are referred to the home page.

The user should be returned to the page they were on after they have accepted/managed cookies

Fixes LPAL-13

## Approach

Due to tech debt this initial approach to this was scrapped. Which would have been a dynamic redirect after the form was submitted. This has been changed to opening the cookie settings page, via the cookie banner. In a new page, where after submitting the form a temp cookie is created with a TTL of 30 seconds & subsequent JS to detect & close the page.

## Learning

## Checklist


service-front/module/Application/view/layout/twig/layout.twig
```diff
@@ -152,7 +152,7 @@
     </script>
 
     <script src="/assets/v2/js/libs/jquery.min.js"></script>
-    <script src="{{ StaticAssetPath( '/assets/v2/js/application.js', { 'minify':true } ) }}"></script>
```
townxelliot: Should minify be re-enabled now? (assuming you switched it off for debugging purposes)

nickdavis2001 (approved):
Approved. Seems like a decent job under the circumstances, I slightly worry about the timeout to close it and lack of a test (although to be fair thats difficult to implement right now). Would be good to get @williamfalconeruk to take a look too.

service-front/public/assets/v2/js/application.js
```diff
@@ -25666,6 +25666,12 @@ this["lpa"]["templates"]["shared.loading-popup"] = Handlebars.template({"compile
                 window.GOVUK.cookie('cookie_policy') || window.GOVUK.setDefaultConsentCookie()
             }
 
+            if (this.isInCookiesPage() && !this.isInIframe() && 'true' === window.GOVUK.cookie('submitted_cookie_page')) {
```
townxelliot, line 25669: I'm not sure what the purpose of this is. Could you please explain? Thanks.

Raffers: The conditional or the timeout below?
The conditional is set as the Cookie Consent section is checked on every page loaded & as I've created a very short-lived cookie to check if cookie options have been submitted. As it was causing further pain to see if we can just add a param to the URL.

The time out I've included as unless I included this it wouldn't run correctly locally.

service-front/module/Application/view/layout/twig/layout.twig
Raffers: Yer, I've reverted this back.

williamfalconeruk (approved):

PR #196, opened 2020-11-27
Lpal 129 vuln fix for pdf container

williamfalconeruk:
## Purpose
We are seeing a number of vulnerabilities being picked up in the pdf (previously masked by an unknown CVE) container scan. the safest option here is to move to an alpine linux container, which has less in it. This will mean a number of changes as PDFTK is not supported as a package since 3.8 of alpine. 
However there is a maintainer of a java ported version, which we are now using.
 
This appears to be the recommended approach..

Fixes LPAL-129

## Approach
- Update dockerfile to Alpine
- Replace install libraries and include those needed for php, xdebug and bash
- Add install of openjdk8-jre to support the java port
- pull version 3.2.1 of the pdftk-java jar
- include executable shell script to sit over the top of the jar file

## Learning
Context:
- Removal of support for older PDFTK: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=892539
- stackoverflow research (relates to ubuntu, but gave pointers) : https://askubuntu.com/questions/1028522/how-can-i-install-pdftk-in-ubuntu-18-04-and-later#new-answer
- deprecation issue raised in Alpine: https://gitlab.alpinelinux.org/alpine/aports/-/issues/10136 

Sources:
- pdftk-java gitlab repo: https://gitlab.com/pdftk-java/pdftk
- pdftk Docker file definition: https://github.com/clevyr/docker-pdftk-java  

Note - PDFTK is pinned to 3.2.1 as the last stable release from the above gitlab repo. our internal channels are subscribed to the rss feed for this, so we can monitor for changes. 

in addition the pdftk-java repo maintainer is attempting to coordinate with the docker file repo maintainer to keep versions up-to-date, so we could use that [directly](https://github.com/clevyr/docker-pdftk-java/blob/master/README.md) instead in future.

## Checklist

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -321,7 +321,7 @@ orbs:
                 if << parameters.unit_test >> ; then
                   docker run -d --env AWS_ACCESS_KEY_ID='-' --env AWS_SECRET_ACCESS_KEY='-' --name tests << parameters.ecr_url >>/<< parameters.ecr_repository_name_prefix >>_app:latest
                   docker exec tests docker-php-ext-enable xdebug
-                  docker exec tests /app/vendor/bin/phpunit --d memory_limit=256M
+                  docker exec tests /app/vendor/bin/phpunit -d memory_limit=256M
                 else
                   echo "phpunit tests not required for this service"
                 fi
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -321,7 +321,7 @@ orbs:
                 if << parameters.unit_test >> ; then
                   docker run -d --env AWS_ACCESS_KEY_ID='-' --env AWS_SECRET_ACCESS_KEY='-' --name tests << parameters.ecr_url >>/<< parameters.ecr_repository_name_prefix >>_app:latest
                   docker exec tests docker-php-ext-enable xdebug
-                  docker exec tests /app/vendor/bin/phpunit
+                  docker exec tests /app/vendor/bin/phpunit --d memory_limit=256M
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -321,7 +321,7 @@ orbs:
                 if << parameters.unit_test >> ; then
                   docker run -d --env AWS_ACCESS_KEY_ID='-' --env AWS_SECRET_ACCESS_KEY='-' --name tests << parameters.ecr_url >>/<< parameters.ecr_repository_name_prefix >>_app:latest
                   docker exec tests docker-php-ext-enable xdebug
-                  docker exec tests /app/vendor/bin/phpunit
+                  docker exec tests /app/vendor/bin/phpunit -d memory_limit=256M
```
charlesmmarshall, line 324: Hey - do we want the memory limit change for the phpunit tests in this pr?

williamfalconeruk: i needed it in to test the build through, can remove this after.

charlesmmarshall: 👍

charlesmmarshall (dismissed):

williamfalconeruk:
Slack message:
Screenshot 2020-11-27 at 13 49 12
ECR:
Screenshot 2020-11-27 at 13 51 05

Look no vulnerabilities!

charlesmmarshall (approved):

Raffers (approved):

PR #198, opened 2020-12-01
more Cypress steps/features to replace Casper

nickdavis2001:
## Purpose
_Adding more steps and features to replace Casper with Cypress _

## Approach

Re-implement existing Casper tests because although these are sub-optimal, it ensures we still have the same level of testing as a starting point. It will then be easy to restructure this as a next step.

- Rewrite S3Monitor in python to keep Dockerfile much simpler and avoid PHP dependencies. 
- New start.sh for kicking this off.
- Re-implement Casper 01-Homepage tests in Cypress as Homepage.feature and step files. Currently, 06-noCookies is not re-implemented, because cypress doesn't support disabling cookies - see https://github.com/cypress-io/cypress/issues/5560.
- Re-implement Casper 02-Signup tests in Cypress as Signup.feature and step files. Currently, logout test is commented out in feature file as it doesn't work yet.
- Re-implement Casper 03-Authentication tests in Cypress as Authentication.feature and step files. 
- Re-implement Casper 04-PasswordReset and 35-ChangePassword tests in Cypress as Password.feature and step files. 
- Re-implement Casper 99-Feedback tests in Cypress as Feedback.feature and step files. 

Casper tests left to do, in future PR(s) are , get the logout part of Signup working, do from Casper 05-CreatePfLPA (this is started, in CreatePfLPA.feature), 15-CreateHwLpa, 25-CreatePfLPAClone

## Learning

Found some issues including :

- Title dropdown (Mr/Mrs etc) when filling out personal details, after failing validation, wrongly becomes a textbox . https://opgtransform.atlassian.net/browse/LPAL-232
- certain attribute fields in twigs get overwritten by PHP, and it all ultimately gets overwritten by javascript. https://opgtransform.atlassian.net/browse/LPAL-234
- Changing a password to something invalid, after having already changed it with a forgot password link, is broken : https://opgtransform.atlassian.net/browse/LPAL-261
- Some form fields may be hidden when they shouldn't be (not currently tested properly - casper filled them in whether hidden or not, for now cypress does the same, needs revisiting). The extent to this i:e how many pages affected, is currently unknown.
- The only emails tested by Casper are Activation, and Password reset for existing user. Numerous other emails sent by the system aren't currently tested

## Checklist


cypress/S3Monitor.py
```diff
@@ -0,0 +1,131 @@
...
+
+# Extract the plus part from emails of the form:
+# basename+pluspart@example.com
+def getPlusPartFromEmailAddress(email):
+    plusPos = email.find('+')
+    atPos = email.find('@')
+    userIdLength = atPos - plusPos - 1
+    userId = email[plusPos + 1:atPos]
+    return userId
+
```
townxelliot, line 46: Suggest using a regex and groups instead of character positions, e.g.

```
import re

def getPlusPartFromEmailAddress(email):
    match = re.search("[^\+]\+(.+)@", email)
    if match is None:
        return ""
    return match.group(1)
```

nickdavis2001: Thanks that works. :)  I had just replicated the slightly clumsy way done in the original PHP, but this is nicer!

cypress/S3Monitor.py
```diff
@@ -0,0 +1,130 @@
...
+        result = s3Client.get_object(Bucket=mailbox_bucket,Key=s3Key)
+        bodyContent = quopri.decodestring(result["Body"].read()).decode('latin-1')
+        #print(f'Parsing {s3Key}')
+        parse_email(bodyContent, s3Key)
+
+def monitor_bucket(s3Client):
+    seenkeys = []
+
+    while True:
+        #print('Checking S3')
```
Raffers, line 111: This comment & the below on line 119/120, are they still needed?

cypress/S3Monitor.py
```diff
@@ -0,0 +1,130 @@
...
+
+    if re.search(reset_password_no_account_subject, bodyContent) is not None:
+        print("Found Password reset for a non-existent account. This shouldn't happen during tests, one explanation can be running password reset test before test that signs user up")
+        return write_unrecognized_file(s3Key,bodyContent,'noaccountpasswordreset')
+
+    # handle other emails. Ultimately, we should be testing these other emails as well
+    print("Found an email that is not an Activate or Password reset. Don't know what to do with it")
+    write_unrecognized_file(s3Key,bodyContent,'unrecognized')
+
+def write_unrecognized_file(s3Key,bodyContent, filePrefix):
```
Raffers, line 94: The spacing here & on process_bucket_object

cypress/S3Monitor.py
```diff
@@ -0,0 +1,130 @@
...
+    if re.search(reset_password_subject, bodyContent) is not None:
+        return parseBody(bodyContent, reset_password_subject, 'passwordreset', 'forgot-password\/reset')
+
+    if re.search(reset_password_no_account_subject, bodyContent) is not None:
+        print("Found Password reset for a non-existent account. This shouldn't happen during tests, one explanation can be running password reset test before test that signs user up")
+        return write_unrecognized_file(s3Key,bodyContent,'noaccountpasswordreset')
+
+    # handle other emails. Ultimately, we should be testing these other emails as well
+    print("Found an email that is not an Activate or Password reset. Don't know what to do with it")
+    write_unrecognized_file(s3Key,bodyContent,'unrecognized')
```
Raffers, line 92: You've got a lot of spacing consistency issues in this file in the code & function def's

cypress/S3Monitor.py
```diff
@@ -0,0 +1,130 @@
...
+            emailFile.close()
+    else:
+        print(f'Message: {subject} does not match regex {regex}') 
+        print('----------------------------------------------------------------------------------')
+
+def parse_email(bodyContent, s3Key):
+    activate_subject = 'Activate your lasting power of attorney account'
+    reset_password_subject = 'Password reset request'
+    reset_password_no_account_subject = 'Request to reset password'  # this is the title for password resets where the account doesn't exist yet. We may need to test this too ultimately
+    if re.search(activate_subject, bodyContent) is not None:
```
Raffers, line 80: A new line so this is more readable?

cypress/S3Monitor.py
```diff
@@ -0,0 +1,130 @@
...
+    if re.search(reset_password_no_account_subject, bodyContent) is not None:
+        print("Found Password reset for a non-existent account. This shouldn't happen during tests, one explanation can be running password reset test before test that signs user up")
+        return write_unrecognized_file(s3Key,bodyContent,'noaccountpasswordreset')
+
+    # handle other emails. Ultimately, we should be testing these other emails as well
+    print("Found an email that is not an Activate or Password reset. Don't know what to do with it")
+    write_unrecognized_file(s3Key,bodyContent,'unrecognized')
+
+def write_unrecognized_file(s3Key,bodyContent, filePrefix):
+        fileSuffix = s3Key[s3Key.rfind('/')+1:]
```
Raffers, line 95: You've put other bits like this into a separate single letter var before doing any sort of slice. I prefer doing it this way, but for consistency which one is to be followed?

cypress/S3Monitor.py
nickdavis2001: fixed a bunch now

cypress/S3Monitor.py
nickdavis2001: not strictly necessary. now removed for tidiness

cypress/S3Monitor.py
nickdavis2001: comment now moved to clearer place

cypress/S3Monitor.py
nickdavis2001: Generally I put it in another var where I feel its needed for clarity. Didn't feel like that was necessary in this case, and don't feel like that needs to be totally consistent across the code.

Raffers (dismissed):

cypress/S3Monitor.py
```diff
@@ -0,0 +1,127 @@
...
+        activationLink = bodyContent[s:e]
+        print(f'{ thetype } link { activationLink }')
+
+        emailRegex = 'To: (.+)'
+
+        emailMatch = re.search(emailRegex, bodyContent)
+        if emailMatch is not None:
+            es = emailMatch.start()
+            ee = emailMatch.end()
+            toEmail = bodyContent[es:ee]
```
townxelliot, line 63: If you just want the email address, you could do `toEmail = emailMatch.group(1)` here, I think (no need for ee and es).

nickdavis2001: Sure, that's more elegant :) just needed the email address, so , now done the emailMatch.group(1).  Looks to work fine.

townxelliot:
@nickdavis2001 I notice several warnings when I run this locally:

```
npm WARN deprecated cucumber@4.2.1: The npm package has moved to @cucumber/cucumber
npm WARN deprecated debug@4.1.1: Debug versions >=3.2.0 <3.2.7 || >=4 <4.3.1 have a low-severity ReDos regression when used in a Node.js environment. It is recommended you upgrade to 3.2.7 or 4.3.1. (https://github.com/visionmedia/debug/issues/797)
npm WARN deprecated chokidar@2.1.8: Chokidar 2 will break on node v14+. Upgrade to chokidar 3 with 15x less dependencies.
npm WARN deprecated fsevents@1.2.13: fsevents 1 will break on node v14+ and could be using insecure binaries. Upgrade to fsevents 2.
npm WARN deprecated urix@0.1.0: Please see https://github.com/lydell/urix#deprecated
npm WARN deprecated resolve-url@0.2.1: https://github.com/lydell/resolve-url#deprecated
npm WARN deprecated core-js@2.6.12: core-js@<3 is no longer maintained and not recommended for usage due to the number of issues. Please, upgrade your dependencies to the actual version of core-js@3.
```

I think we can still merge, but once in should raise another issue to upgrade the cypress container (I think that's what's producing this) to remove these errors.

Makefile
```diff
@@ -154,7 +154,10 @@ integration-api-local:
 .PHONY: cypress-local
 cypress-local:
 	docker build -f ./cypress/Dockerfile  -t cypress:latest .; \
-	docker run -it -e "CYPRESS_baseUrl=https://localhost:7002" --network="host" --rm cypress:latest cypress run --spec cypress/integration/BasicLogin.feature
+	#aws-vault exec moj-lpa-dev -- docker run -it -e AWS_ACCESS_KEY_ID -e AWS_SECRET_ACCESS_KEY -e AWS_SESSION_TOKEN -e "CYPRESS_baseUrl=https://localhost:7002" --entrypoint ./cypress/start.sh --network="host" --rm cypress:latest run
```
townxelliot, line 157: Please remove spurious makefile lines and statements

cypress/Dockerfile
```diff
@@ -2,9 +2,17 @@ FROM cypress/included:5.2.0
 
 WORKDIR /app
 
-RUN npm install --save-dev cypress-axe@0.9.1 axe-core cypress-cucumber-preprocessor cosmiconfig
+RUN apt-get update && \
+  apt-get install -y python3-pip 
 
+RUN pip3 install boto3
+
+RUN npm install --save-dev cypress-axe@0.9.1 axe-core cypress-cucumber-preprocessor cosmiconfig cypress-tags
```
townxelliot, line 10: cosmiconfig is not required here as far as I can tell; cypress-tags is a dependency of cypress-cucumber-preprocessor so we could remove that (as we're relying on the cypress-tags script wrapper which is in cypress-cucumber-preprocessor, not cypress-tags directly)

Makefile
```diff
@@ -154,7 +154,10 @@ integration-api-local:
...
+#	aws-vault exec moj-lpa-dev -- docker run -it -e AWS_ACCESS_KEY_ID -e AWS_SECRET_ACCESS_KEY -e AWS_SESSION_TOKEN -e "CYPRESS_baseUrl=https://localhost:7002" --entrypoint bash --network="host" --rm cypress:latest 
+# attempt to use cypress-tags under development. does not currently work , uncomment to make work
+	aws-vault exec moj-lpa-dev -- docker run -it -e AWS_ACCESS_KEY_ID -e AWS_SECRET_ACCESS_KEY -e AWS_SESSION_TOKEN -e "CYPRESS_baseUrl=https://localhost:7002" --entrypoint ./node_modules/.bin/cypress-tags --network="host" --rm cypress:latest run -e TAGS='not @SignUp'
```
townxelliot, line 160: As the CreatePfLpa.feature tests fail, I suggest we add that to the `TAGS` specification so it isn't run; otherwise the cypress tests fail and will break the build (unless you're planning on fixing it before we merge this PR).

cypress/integration/Password.feature
```diff
@@ -0,0 +1,51 @@
...
+        Then I am taken to the type or dashboard page
+        # change password back to old one. This wasn't in the original casper tests, but ensures this feature doesn't have any side effects 
+        When I visit link containing "Your details"
+        Then I am taken to "/user/about-you"
+        When I visit link containing "Change Password"
+        Then I am taken to "/user/change-password"
+        When I change password back to my old one
+        And I see "Your new password has been saved" in the page text
+
+
```
townxelliot, line 51: These tests fail for me with:

```
Running:  Password.feature                                                                (5 of 5)


  Password
    1) Change existing password with invalid details
    2) Reset Password using email link


  0 passing (2m)
  2 failing

  1) Password
       Change existing password with invalid details:
     AssertionError: Timed out retrying: Expected to find content: 'Your details' but never did.
      at Context.eval (https://localhost:7002/__cypress/tests?p=cypress/integration/Password.feature:344:6)
      at Context.resolveAndRunStepDefinition (https://localhost:7002/__cypress/tests?p=cypress/integration/Password.feature:11940:9)
      at Context.eval (https://localhost:7002/__cypress/tests?p=cypress/integration/Password.feature:11281:35)

  2) Password
       Reset Password using email link:
     AssertionError: Timed out retrying: `cy.readFile("cypress/activation_emails/undefined.passwordreset")` failed because the file does not exist at the following path:

`/app/cypress/activation_emails/undefined.passwordreset`

https://on.cypress.io/readfile
      at openEmailAndVisitLink (https://localhost:7002/__cypress/tests?p=cypress/integration/Password.feature:81:6)
      at Context.eval (https://localhost:7002/__cypress/tests?p=cypress/integration/Password.feature:75:3)
      at Context.resolveAndRunStepDefinition (https://localhost:7002/__cypress/tests?p=cypress/integration/Password.feature:11940:9)
      at Context.eval (https://localhost:7002/__cypress/tests?p=cypress/integration/Password.feature:11281:35)
```

Any ideas?

townxelliot (changes requested):
Thanks @nickdavis2001. I tried the test locally and spotted a few things which I've commented on here. I will try doing a full reset and running them again in case it's something in my environment (stale containers etc.).

cypress/integration/Password.feature
nickdavis2001: You have to run Password.feature after Signup.feature, during that run of the docker container. It resets the password for a user that has signed up. This is what the Casper tests did. Ultimately, as previously discussed as a group, we ought to have some code which, if it fails to find the signed up user, goes and makes that on the fly, although we ought to retain as part of the tests in CI, the scenario of a user being signed up via the website and everything being tested again after that.

cypress/Dockerfile
nickdavis2001: yeah cosmiconfig was legacy from getting cypress-axe and cypress-cucumber to play together and that is true cypress-tags gets pulled in anyway

Makefile
nickdavis2001: I've removed the accordion check for now. This will mean this test always passes. We can work out later how to be sure we only check the accordion when it is meant to be there. Regarding breaking the build, a failure in Cypress tests at the moment gets swallowed. We could now perhaps switch that so it does indeed break the build. However... the problem then is a11y tests would also break it. So we need a way first to get it to do conditional fails i:e we swallow a11y errors but not others. I'll make a story for that.

Makefile
nickdavis2001: Done.

cypress/integration/Password.feature
townxelliot: Thanks for clarifying. Does this mean we also need to switch these tests off if signup is off?

cypress/start.sh
```diff
@@ -0,0 +1,46 @@
...
+    echo "Running Cypress GUI"
+# pass supplied args to cypress
+    CYPRESS_CMD="cypress $@"
+    echo $CYPRESS_CMD
+    $CYPRESS_CMD
+else
+    # Its headless (used in CircleCI) so run the signup test first followed by all others
+    echo "Running Cypress headless"
+    ./node_modules/.bin/cypress-tags run -e TAGS='@SignUp' 
+	./node_modules/.bin/cypress-tags run -e TAGS='not @SignUp'
```
townxelliot, line 33: indent is out here

Makefile
```diff
@@ -154,7 +154,11 @@ integration-api-local:
 .PHONY: cypress-local
 cypress-local:
 	docker build -f ./cypress/Dockerfile  -t cypress:latest .; \
-	docker run -it -e "CYPRESS_baseUrl=https://localhost:7002" --network="host" --rm cypress:latest cypress run --spec cypress/integration/BasicLogin.feature
+	aws-vault exec moj-lpa-dev -- docker run -it -e AWS_ACCESS_KEY_ID -e AWS_SECRET_ACCESS_KEY -e AWS_SESSION_TOKEN -e "CYPRESS_baseUrl=https://localhost:7002" --entrypoint ./node_modules/.bin/cypress-tags --network="host" --rm cypress:latest run -e TAGS='not @SignUp'
```
townxelliot, line 157: I would remove `TAGS` from this command line as it's ignored in start.sh anyway.

cypress/integration/Password.feature
nickdavis2001: In CI, we explicitly run the Signup tests (tests tagged Signup) followed by all tests not tagged Signup. When running locally through the GUI, you can easily do that by hand. I have now changed the cypress-local target in Makefile, so it does exactly what CI does, which sorts this.

Makefile
nickdavis2001: now does the same way as CI

townxelliot (approved):

PR #199, opened 2020-12-01
LPAL-23 - I want to see that relevant guidance on the LPA type page

Raffers:
## Purpose
As a first time user of the service, when I complete the "my details" page and then directed to the LPA type page, or log in for the first time after having only previously completed ‘my details’, I want to see the page from the top.

The page currently lands/anchors to the ‘What type of LPA’ question header and making it unclear to users that there is content with links above.

Fixes LPAL-23

## Approach

Removes the element that was forcing the focus change via JS

## Learning

## Checklist


williamfalconeruk (approved):

service-front/module/Application/src/Model/Service/Session/SessionManager.php
```diff
@@ -6,6 +6,11 @@
 
 class SessionManager extends LaminasSessionManager {
 
+    /**
+     * @var string $lastMatchedRoute
+     */
+    private $lastMatchedRoute;
```
Raffers, line 12: I feel like this should be stored elsewhere.

service-front/module/Application/src/Module.php
```diff
@@ -10,8 +10,6 @@
 use Application\Model\Service\System\DynamoCronLock;
 use Alphagov\Pay\Client as GovPayClient;
 use Aws\DynamoDb\DynamoDbClient;
-use Opg\Lpa\Logger\LoggerTrait;
-use TheIconic\Tracking\GoogleAnalytics\Analytics;
```
williamfalconeruk, line 14: should this be removed? see line 145 of this file

service-front/module/Application/src/View/Helper/RouteName.php
```diff
@@ -2,32 +2,29 @@
...
     public function __invoke()
     {
-        $routeName = '';
-
-        if ($this->routeMatch) {
-            $routeName = $this->routeMatch->getMatchedRouteName();
-        }
-
-        return $routeName;
+        return $this->routes;
```
townxelliot, line 28: Just out of interest, is `__invoke()` ever called? I imagine instances of this will come from `RouteNameFactory`, which calls the constructor for this class with `new`.

service-front/module/Application/view/layout/twig/accordion-layout.twig
```diff
@@ -14,11 +14,12 @@
     <ul class="accordion">
 
         {% set accordionTop = accordion( lpa ).top() %}
+        {% set route = routeName() %}
 
         {% include 'layout/partials/accordion/accordion.twig' with { 'items' : accordionTop, 'lpa':lpa } %}
 
-        {% if (accordionTop|last).routeName != 'review-link' %}
-            <li aria-hidden="true"><a id="current" class="accordion-current"></a></li>
+        {% if route['previous'] != 'user/about-you' and route['previous'] != 'user/dashboard' and (accordionTop|last).routeName != 'review-link' %}
```
townxelliot, line 21: Does modifying `routeName()` have any effect on what gets set in `(accordionTop|last).routeName`? Does that need to be modified so that it uses the `current` property from the routeName or is it unaffected?

townxelliot (changes requested):
It looks like this no longer has anything to do with JavaScript, and is now about modifying the twig template so it doesn't write the HTML which would enable the accordion to focus on it (albeit with JS). Is that how this works?

This also changes the accordion focus for all users, not just new users or users who have signed in for the first time. Is that the case? I don't see it as a problem, but the original Jira issue specifically calls out new users/first time sign in and I wanted to make sure this is the behaviour we want for all users.

service-front/module/Application/src/Module.php
```diff
@@ -156,6 +156,27 @@ public function getServiceConfig()
...
+                    $sessionDetails = new Container('SessionDetails');
+
+                    // breadcrumb so we can determine user's last visited route.
+                    $sessionDetails->currentRoute = $sm->get('Application')->getMvcEvent()->getRouteMatch()->getMatchedRouteName();
+
+                    if (!isset($sessionDetails->previousRoute)) {
+                        $sessionDetails->previousRoute = $sessionDetails->currentRoute;
+                    }
+
+                    if ($sessionDetails->routeStore !== $sessionDetails->previousRoute) {
```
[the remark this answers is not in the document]
Raffers: No, it's purely used as a switch. So we're not just overwriting the previous with the current route. I thought it better to do it this way, with a comparison. Rather than use a bool to do the switching.

service-front/module/Application/src/View/Helper/RouteName.php
Raffers: It's invoked when `RouteName` is used within a twig template.

service-front/module/Application/view/layout/twig/accordion-layout.twig
Raffers: They're entirely separate. I can't remember the class right now, but when I was initially looking at this it's a property of that class. 

Although we could just replace that now & use `routeName()`, but I thought it was a bit of, a can of worms. With how the track the path when working with the LPA itself.

service-front/module/Application/src/ControllerFactory/ControllerAbstractFactory.php
```diff
@@ -133,18 +133,22 @@ public function __invoke(ContainerInterface $container, $requestedName, array $o
         $controllerName = $this->getControllerName($requestedName);
 
         $formElementManager = $container->get('FormElementManager');
+        // Var not used within class, but required to populate twig helper function within templates.
+        $sessionDetails = $container->get('PersistentSessionDetails');
```
townxelliot, line 137: I'm not sure about this comment: how is this used in the twig helper? Is it acting somehow like a global variable?

townxelliot (changes requested):

service-front/module/Application/src/ControllerFactory/ControllerAbstractFactory.php
Raffers: Unless the container I've constructed is called, there won't be anything contained within it. I've had to place this low enough so it's getting the right routing information, but high enough that it doesn't interfere with anything. My comment doesn't explain it very well. It's a little hacky, to be honest.

townxelliot: Right, so how is `$sessionDetails` called by the template? Are you effectively just initialising the PersistentSessionDetails (in which case the variable is not necessary)?

Raffers: Oh, it's just initiated. I put it in a var out of habit more than anything, I'll change that.

Also, in reply to your other comment, as I can't reply to that directly. The acordin is loaded, I think, on every page. But only takes effect in a few circumstances. Mainly when it's used in either creation of the LPA, so it can display additional guidance or when they first start the LPA journey to skip the guidance at the top of the page regarding LPA types.

townxelliot (approved):

PR #200, opened 2020-12-02
LPAL-206 Further functional test and local run improvements

townxelliot:
## Purpose

Work to get functional tests running locally.

Fixes LPAL-206.

## Approach

* Improve start up sequence of docker containers esp. so that SQL going to postgres from different places doesn't cross beams:
  * api app waits before applying migrations
  * seeding waits for tables created by api app migrations before inserting seed data
* Improve tests and scripts so they don't have unnecessary waits:
  * casperjs tests use waitForSelector() rather than wait() for set time period
  * Waits in seeding script replaced with polling for endpoints/db
* Reduce weight of docker container for seeding (doesn't require Python only Alpine)

## Learning

Functional tests are fragile and inconsistent, and PDF generation hangs for a very long time (on my local setup, I'm still unable to generate LP3 PDF in under 8 minutes).

## Checklist

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -164,6 +164,7 @@ services:
       - dynamodb
       - localstack
       - local-config
+      - postgres
     environment:
       OPG_LPA_STACK_NAME: "local"
       OPG_LPA_STACK_ENVIRONMENT: "dev"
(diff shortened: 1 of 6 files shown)


nickdavis2001 (approved):
🏆   Very helpful to all current and future Make developers :)

williamfalconeruk (approved):

PR #203, opened 2020-12-10
Lpal 193 Live url upgrades and environment migration to management account(DO NOT MERGE UNTIL READY)

williamfalconeruk:
## Purpose

- To add new public facing dns entry to the management account, so that non-production environments now support this.
- To add the public facing "live" style url to allow our test suite to use this instead of the internal domain. 
- Make this new url available in CircleCI and post on Slack as part of dev environments build
- Assign front Application Load Balancer (ALB) to the new dns entry and live certificate for account.

See #202 also, but further cert changes will happen after we change over Route53 Zone name servers.

Partially Fixes LPAL-193 

## Approach

- Terraform changes
- Circleci and script changes
- Remove hard coded locals for domains and calculate them instead based on environment.

## Learning
- AWS Application Load Balancers (ALB): https://docs.aws.amazon.com/elasticloadbalancing/latest/application/introduction.html
- AWS Load balancer Listeners for Targetgroups: https://docs.aws.amazon.com/elasticloadbalancing/latest/application/load-balancer-listeners.html
- AWS Target Groups: https://docs.aws.amazon.com/elasticloadbalancing/latest/application/load-balancer-target-groups.html

## Checklist

* [ ] ~~I have added tests to prove my work~~/ I have planned my changes in terraform


terraform/environment/load_balancer_front.tf
```diff
@@ -130,15 +129,15 @@ resource "aws_lb_listener_rule" "www_redirect" {
     type = "redirect"
 
     redirect {
-      host        = "www.lastingpowerofattorney.service.gov.uk"
+      host        = "www.${local.dns_namespace_env}lastingpowerofattorney.service.gov.uk"
```
Raffers, line 132: With the `dns_namespace_env` var being populated with pre-prod URL will `www` addressing still work?

williamfalconeruk: i'm about to find out :)

charlesmmarshall (approved):

PR #205, opened 2020-12-15
Mock X-Amzn-Trace-Id on requests and forward to *-web containers

townxelliot:
## Purpose

Provide X-Amzn-Trace-Id header from *-ssl containers to
*-web containers.

At this point, we should be able to trace requests from the load
balancer to the *-web containers (or from the mock *-ssl load
balancers in dev to the *-web containers).

Part of LPAL-9

Fixes LPAL-30 (https://opgtransform.atlassian.net/browse/LPAL-30)

## Approach

* Use a tweaked docker-ssl-proxy image, with config modified
to include an X-Amzn-Trace-Id header on proxied requests.
* Modify *-web nginx config to forward the X-Amzn-Trace-Id
to the FastCGI handler.
* Turn on custom logging in *-web nginx containers to include
the trace ID.

## Learning

Logging in LPA Live is inconsistent, which I'll attempt to address in future PRs.

## Checklist


williamfalconeruk (dismissed):

townxelliot:
@williamfalconeruk Thinking about whether I should revisit this and move to JSON logging as in Sirius, while we have the chance. Just need to confirm that this won't scupper any other systems we're using for analysing logs. Any thoughts?

townxelliot:
@williamfalconeruk I updated this PR with some additional commits; this now produces JSON logs and removes logging of static assets. I'm not sure if we would prefer to log static assets, just not as JSON.

townxelliot:
Discussing with @williamfalconeruk whether we should retain some logging for static assets, and whether those assets should be logged in NCSA format rather than JSON. This would mean a log with entries in two formats, though.

williamfalconeruk (approved):

nickdavis2001 (approved):

williamfalconeruk (approved):

PR #215, opened 2021-01-06
LPAL-31 and LPAL-32 trace id logging

townxelliot:
## Purpose

Rework logging across front-web, front-app, api-web and api-app so that we pass the X-Trace-Id header from incoming requests to dependent services and log it in PHP logs where appropriate.

Fixes LPAL-31, LPAL-32

## Approach

- Use custom JSON-formatted logging for PHP logs (front-app, api-app).
- Modify API client factory to inject incoming X-Trace-Id header into outgoing requests.
- Add an MvcEvent listener which logs any controller exceptions with their trace ID.

Note that there is a lot more we could do here to improve logging. This PR simply standardises the PHP logging which is already there to produce JSON containing the trace ID.

WARNING: This changes the log format for all services in the Make stack. If there are any reports which depend on log format, they will need to be modified for the new JSON log style.

## Learning

We need to refactor the code base so we can share common code without having to create a repository for it.

## Checklist


TECH_DEBT.md
```diff
@@ -0,0 +1,10 @@
+# Technical debt to pay back
+
+* We added a new custom Logger to service-front, to enable us to log as JSON and log MVC error events in more detail. Consequently, we should remove opg-lpa-logger from service-front/composer.json, as it's no longer used. However, removing this and updating composer.lock causes the app to break. Need to investigate why that is and upgrade composer packages.
```
Raffers, line 3: Probably a daft question, but were you manually doing this, or did you try removing it via composer cli?

townxelliot:
@Raffers thanks for your comments; I force pushed onto this branch which lost your comment, unfortunately. But no, I hadn't thought of that, but will use that approach to remove the OPG LPA logger from composer.json.

TECH_DEBT.md
williamfalconeruk, line 3: Will we raise tickets for these tech debt items?

townxelliot: I may just do that and remove this document actually. I already raised the issue for this.

service-front/module/Application/src/Controller/Console/DynamoDbController.php
```diff
@@ -4,13 +4,10 @@
 
 use Aws\DynamoDb\Exception\DynamoDbException;
 use Aws\Sdk;
-use Opg\Lpa\Logger\LoggerTrait;
```
williamfalconeruk, line 7: noticed that this had been removed, but not replaced. having looked at the file , it has a bunch of echo statements. wondering if we need to do something here (could be a further TD ticket...)

service-front/module/Application/src/Model/Service/Analytics/GoogleAnalyticsService.php
```diff
@@ -4,13 +4,10 @@
 
 use Application\Model\Service\AbstractService;
 use Exception;
-use Opg\Lpa\Logger\LoggerTrait;
```
williamfalconeruk, line 7: just checking if this removal, non replacement was intentional.

service-front/module/Application/src/Model/Service/Lpa/Application.php
```diff
@@ -22,13 +22,11 @@
 use Opg\Lpa\DataModel\Lpa\Payment\Payment;
 use Opg\Lpa\DataModel\WhoAreYou\WhoAreYou;
 use ArrayObject;
-use Opg\Lpa\Logger\LoggerTrait;
```
williamfalconeruk, line 25: just checking if removal, non replacement was intentional

service-front/module/Application/src/Module.php
```diff
@@ -10,7 +10,6 @@
 use Application\Model\Service\System\DynamoCronLock;
 use Alphagov\Pay\Client as GovPayClient;
 use Aws\DynamoDb\DynamoDbClient;
-use Opg\Lpa\Logger\LoggerTrait;
```
williamfalconeruk, line 13: just checking this was an intentional removal with no replace.

williamfalconeruk (commented):
left a bunch of questions - nothing too onerous - looks good though!

service-front/module/Application/src/Controller/Console/DynamoDbController.php
townxelliot: It's a console app, so rather than using the logger, it writes direct to stdout. Not sure if we need to do anything more than that as we rarely use this anyway (do we?).

service-front/module/Application/src/Model/Service/Analytics/GoogleAnalyticsService.php
townxelliot: Good catch @williamfalconeruk this does need to reference the new logger class.

service-front/module/Application/src/Model/Service/Lpa/Application.php
townxelliot: logger is now inherited from AbstractService.

TECH_DEBT.md
townxelliot: I added a commit to remove this document and added issues for the tech debt.

Raffers (approved):

Raffers:
> @Raffers thanks for your comments; I force pushed onto this branch which lost your comment, unfortunately. But no, I hadn't thought of that, but will use that approach to remove the OPG LPA logger from composer.json.

I actually deleted it, as the comment directly below said that you had tried other composer CLI commands. So I had assumed you would have tried composer dump-autoload along with it.

williamfalconeruk (approved):

PR #218, opened 2021-01-11
LPAL-133 and LPAL-172 add maintenance mode to lpa

williamfalconeruk:
## Purpose
To move the maintenance page to be inline with other LPA services, and how they operate. 

Fixes LPAL-133 and LPAL-172

## Approach

- Add script to switch on maintenance mode using AWS SSM parameter store.
- Add Load balancer rule based on AWS SSM Parameter store value.
- Add minified html page and fit content (made similar to use and refunds).
- Runbook instructions.

## Learning
- AWS SSM parameter store: https://docs.aws.amazon.com/systems-manager/latest/userguide/systems-manager-parameter-store.html
## Checklist


williamfalconeruk:
preview of new maintenance page: 
Screenshot 2021-01-11 at 13 42 35

williamfalconeruk:
Updated content
Screenshot 2021-01-12 at 10 02 49

docs/runbooks/maintenance_mode/README.md
```diff
@@ -0,0 +1,56 @@
+# Manage Maintenance Mode
+
+This script will enable or disable maintenance mode for a targeted environment.
+this can be run locally or via a cloud9 instance, dependant on needs.
+
+## Setup - Local Credentials Route
+
+Use this route if you have the required access via aws-vault to make changes to the environment that needs to be put into maintenenance mode.
+
+You will need have to set up assumable roles in aws-vault. Follow the instructions at [setting up credentials](../setting-up-aws-credentials/setting-up-credentials.md)
```
townxelliot, line 10: You will need ~~have~~ to

docs/runbooks/maintenance_mode/README.md
```diff
@@ -0,0 +1,56 @@
...
+
+To turn maintenance mode on (e.g. in preproduction):
+
+```sh
+aws-vault exec moj-lpa-preprod -- ./manage_maintenance.sh \
+  --environment preproduction \
+  --maintenance_mode
+```
+
+to turn maintenece mode off:
```
townxelliot, line 22: maintenance

docs/runbooks/maintenance_mode/README.md
```diff
@@ -0,0 +1,56 @@
+# Manage Maintenance Mode
+
+This script will enable or disable maintenance mode for a targeted environment.
+this can be run locally or via a cloud9 instance, dependant on needs.
```
townxelliot, line 4: ~~t~~This

docs/runbooks/maintenance_mode/README.md
```diff
@@ -0,0 +1,56 @@
...
+
+```sh
+aws-vault exec moj-lpa-preprod -- ./manage_maintenance.sh \
+  --environment preproduction \
+  --disable_maintenance_mode
+```
+
+## Setup - Remote Route
+
+If you dont have this, you can also set up cloud9 on the relevant account in the AWS console.
```
townxelliot, line 32: don't

docs/runbooks/maintenance_mode/manage_maintenance.sh
```diff
@@ -0,0 +1,62 @@
+#!/usr/bin/env bash
```
townxelliot, line 1: Wondering about scripts: should these be mixed in with the runbooks or further up the project?

terraform/environment/load_balancer_front.tf
```diff
@@ -98,6 +98,46 @@ resource "aws_lb_listener_certificate" "front_loadbalancer_live_service_certific
...
+    type = "fixed-response"
+
+    fixed_response {
+      content_type = "text/html"
+      message_body = file("${path.module}/maintenance/maintenance.html")
+      status_code  = "503"
+    }
+  }
+
+  condition {
```
townxelliot, line 128: As I'm not familiar, could you explain how is works at a high level? I can see that you're setting a fixed response on the load balancer, but I'm not sure how setting the maintenance switch tells the site to show the maintenance page. Is it that you're changing the condition on the rule (from the bash script) so that any request with a particular value will trigger the rule that shows the maintenance page?

townxelliot (changes requested):
Looks good, just a few typos and a request for clarification, thanks.

docs/runbooks/maintenance_mode/manage_maintenance.sh
williamfalconeruk: IMO, best to keep these with the runbooks as close to the instructions as possible, for easier access- these tend to be troubleshooting and operational scripts that help remedy a situation. We do have a scripts folder also but these tend to align with the pipeline.

terraform/environment/load_balancer_front.tf
williamfalconeruk: High level is 
- we set up a Load balancer rule that gives a fixed response. 
- in Terraform we ignore condition lifecycle changes, so that we can modify using the script
- the script modifys the rule to enable the fixed response or disables it accordingly, based on the bash script, allowing quick switch over.
I have noticed we might not need the SSM  parameter, i'll test that again without...

docs/runbooks/maintenance_mode/manage_maintenance.sh
williamfalconeruk: Also other teams have adopted a similar pattern.

townxelliot (approved):

PR #219, opened 2021-01-12
LPAL-23 - I want to see that relevant guidance on the LPA type page

Raffers:
## Purpose

As a first time user of the service, when I complete the "my details" page and then directed to the LPA type page, or log in for the first time after having only previously completed ‘my details’, I want to see the page from the top.

The page currently lands/anchors to the ‘What type of LPA’ question header and making it unclear to users that there is content with links above.

Fixes LPAL-23

## Approach

Creates a breadcrumb system that stores the previously accessed route, which is then passed to the RouteName twig helper function

## Learning

## Checklist

### DIFF
--- a/service-front/module/Application/src/Model/Service/Session/PersistentSessionDetails.php
+++ b/service-front/module/Application/src/Model/Service/Session/PersistentSessionDetails.php
@@ -33,10 +33,6 @@ class PersistentSessionDetails {
         $this->sessionDetails->currentRoute = ($this->route instanceof RouteMatch) ?
             $this->route->getMatchedRouteName() :
             '';
-        
-        if (!isset($this->sessionDetails->previousRoute)) {
-            $this->sessionDetails->previousRoute = $this->sessionDetails->currentRoute;
-        }
 
         if ($this->sessionDetails->routeStore !== $this->sessionDetails->previousRoute) {
             $this->sessionDetails->previousRoute = $this->sessionDetails->routeStore;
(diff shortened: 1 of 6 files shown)


service-front/module/Application/src/Model/Service/Session/PersistentSessionDetails.php
```diff
@@ -50,6 +46,6 @@ public function getCurrentRoute(): string {
     }
 
     public function getPreviousRoute(): string {
-        return $this->sessionDetails->previousRoute;
+        return $this->sessionDetails->previousRoute ?? 'home';
```
williamfalconeruk, line 49: i guess it's reasonable to assume home is the default here?

Raffers: Yer, I thought it best to set this to a default of home, with the inconsistency with how the route is populated. I did consider setting it to an empty string.

Raffers: Have had to add another null check on `getCurrentRoute` as this seems to be still being populated with null somehow, despite the earlier check to make sure we have the `routeMatch`

townxelliot: I think we're going to need a cypress test for this sharpish. How confident are you that we don't need it for this PR @Raffers ?

townxelliot: I picked up the cypress ping test and have done a rough draft, so we can put that in as soon as we've got Nick's cypress work in.

townxelliot:
Suggest we ensure we do a manual test of the /ping endpoint once this PR is merged to ensure it's good to go live. Then we can incorporate my automated test once Nick's work is in.

townxelliot (approved):
Looks good, approved.

williamfalconeruk (approved):

PR #221, opened 2021-01-12
LPAL-264 Cypress test for /ping

townxelliot:
## Purpose

Test the /ping endpoint at least returns a 200 and reports that dynamodb is accessible.

Fixes LPAL-264

## Approach

Added a cypress feature file and impl to test ping.

Does not test that the whole page is without errors, as we always get an error for the gateway in the local environment.

## Learning

Thanks to @nickdavis2001's work this was very straightforward to incorporate.

One thing I couldn't figure out was how to show console logs when running cypress tests from the command line. This seems unnecessarily complicated.

## Checklist


Raffers (approved):

nickdavis2001 (approved):
Very nice :)

nickdavis2001:
Hi Elliott, I couldn't get console logs to easily work either, so that's why I do cy.log all over the place. I wonder can we do some setting to redirect that to console when the GUI isn't running. Might look into that.

cypress/integration/common/ping.js
```diff
@@ -0,0 +1,17 @@
+import { Given } from "cypress-cucumber-preprocessor/steps";
+
+Given('I visit the ping page', (path) => {
+    // note that cy.visit will follow redirects, and require the status code to be 2xx after that
+    cy.visit('/ping');
+})
+
+Given('I visit the ping JSON endpoint', (path) => {
+    cy.request('/ping/json').then((response) => {
```
williamfalconeruk, line 9: Just wondering if the `then` bit could be split into a then step in the feature, something along the lines of  `Then I get a valid  Response` or similar.

nickdavis2001: +1  That's a good point . The bit that checks it, should be in a Then step.  Given I visit the Ping JSON endpoint,  Then I get back valid json .   Then. "Then I get back valid json" step could then be reused in many places.

williamfalconeruk: this might be useful as we could split the tests up... https://example.cypress.io/commands/network-requests

williamfalconeruk: (see 3rd example about shared test context)

williamfalconeruk (approved):
left a discussion point, for possible improvement. but otherwise looks good!

cypress/integration/common/ping.js
townxelliot: I added a commit to make this more generic so it could be used for different endpoints. I also added a generic step for testing object properties. This can be used specifically for JSON, but is also general enough to work with other types of object.

williamfalconeruk (dismissed):
This is a thing of beauty, great reusable test code!

townxelliot:
Thanks for the comment @williamfalconeruk, much appreciated :)

cypress/start.sh
```diff
@@ -29,9 +29,8 @@ if [[ $GUI == "true" ]] ; then
 else
     # Its headless (used in CircleCI) so run the signup test first followed by all others
     echo "Running Cypress headless"
-    ./node_modules/.bin/cypress-tags run -e TAGS='@Ping'
-    #./node_modules/.bin/cypress-tags run -e TAGS='@SignUp'
-    #./node_modules/.bin/cypress-tags run -e TAGS='not @SignUp'
+    ./node_modules/.bin/cypress-tags run -e TAGS='@SignUp'
```
williamfalconeruk, line 32: whoops sorry I missed that!

townxelliot:
I realised I'd turned off all the tests except my own! So I've reinstated them. It may require re-approval. Apologies. Next task is to write a parameterised script so I can specify which tests to include in the run.

williamfalconeruk (approved):

PR #227, opened 2021-01-18
LPAL-259 Link accessibility

townxelliot:
## Purpose

Improve link accessibility across the site, verified with cypress tests.

Fixes LPAL-259

## Approach

* Add "opens in new tab" to all target="_blank" links.
* Where the above is repetitive, hide the "opens in new tab" text to all except screen readers and add a notice that the links in the page open in new tabs.
* Change text on "edit" links to "Change" consistently across the site.
* Modify wording on some links so the context is more obvious and they are more "standalone".
* Add cypress steps to check links on a page, and also to verify that notice about hidden links is on the pages where expected (privacy and terms, mostly).
* Fix link colours to design system spec.
* Add rel="noreferrer noopener" to all target="_blank" links, in line with design system spec.

For UAT purposes, this PR touches the following pages:

* Home
* Privacy
* Terms
* Layout (all pages, but visually hidden to most users)
* Accessibility
* Create account
* LPA summary
* LPA complete
* LPA view docs
* Guidance: Process_How to register the LPA.md
* Guidance: About LPAs_Life-sustaining treatment.md
* Cookies

It includes some text changes to clarify link destinations. I made these as I saw fit, but I'm not sure if we have to have such changes cleared by a higher authority. If we do, I will revert them where necessary.

## Learning

n/a

## Checklist


service-front/assets/sass/patterns/_accordion.scss
```diff
@@ -108,7 +108,7 @@
 
         .link-edit {
             display: block;
-            padding: $gutter-half $gutter;
+            padding: $gutter-half $gutter-half;
```
nickdavis2001, line 111: What was the reason for this, does it affect a lot of pages? Presumably yes? Is it at all dangerous in terms of breaking anything else? Hopefully no. ;)

service-front/module/Application/view/application/general/home/accessibility.twig
```diff
@@ -37,7 +37,7 @@
     </ul>
     <h2 class="heading-medium">What to do if you cannot access parts of this service</h2>
     <p>If you cannot use this service, you can use the <a href="https://www.gov.uk/government/publications/make-a-lasting-power-of-attorney">paper version of the LPA</a>. </p>
-    <p>You can email <a href="mailto:customerservices@publicguardian.gov.uk" target="_blank">customerservices@<wbr>publicguardian.gov.uk</a> to get the paper LPA in an alternative format, such as:</p>
+    <p>You can email <a href="mailto:customerservices@publicguardian.gov.uk">customerservices@<wbr>publicguardian.gov.uk</a> to get the paper LPA in an alternative format, such as:</p>
```
nickdavis2001, line 40: I see you removed a target=blank from this mailto.  Are we definitely getting rid of all of those then?  I thought there was some reason why these might still be needed, but maybe not anymore?

service-front/module/Application/view/application/general/home/accessibility.twig
```diff
@@ -47,7 +47,7 @@
     <h2 class="heading-medium">Reporting accessibility problems with this service</h2>
     <p>We’re always looking to improve the accessibility of this service. If you find any problems not listed on this page or think we’re not meeting accessibility requirements, please contact us. </p>
     <h3 class="heading-small">Email</h4>
-    <p><a href="mailto:customerservices@publicguardian.gov.uk" target="_blank">customerservices@<wbr>publicguardian.gov.uk</a></p>
+    <p><a href="mailto:customerservices@publicguardian.gov.uk">customerservices@<wbr>publicguardian.gov.uk</a></p>
```
nickdavis2001, line 50: Same as above. Are we 100% its OK to get rid of these target blanks for mails?

townxelliot: My reasoning is that this is a mailto: link, and would therefore be expected to open either a new window or a separate application. The fact this is sign-posted in the href suggests we don't need to add a target attribute. Also, if we do have a target attribute, we would need to highlight in the text that this link will open a new tab. It won't necessarily, as the user may have an email application which opens on a click of these links. @johnnolan may have some thoughts.

service-front/assets/sass/patterns/_accordion.scss
townxelliot: It affects all accordions for editing. But because they should all be using the same macro, and therefore all get the altered text ("Change" instead of "edit" or "Edit"), they will all need to use this adjusted CSS.

service-front/module/Application/view/application/general/home/accessibility.twig
townxelliot: (See my answer to your next question)

service-front/module/Application/view/application/general/home/accessibility.twig
townxelliot: I've discussed with @johnnolan and he concurs that we can do without target="_blank" on mailto links as the browser/screenreader will interpret the mailto href appropriately for the user.

nickdavis2001 (approved):
lgtm :)

Raffers (approved):

townxelliot:
NB also waiting for Amy to approve text change to home page before I merge

townxelliot:
I'm going to wait for #229 before merging this, as it will conflict with Chris' work.

Raffers (approved):

Raffers (dismissed):

Raffers (approved):

townxelliot:
Please do not merge: still waiting for a UAT pass.

PR #229, opened 2021-01-20
LPAL-250 and LPAL-246 - Heading tag changes & keyboard navigation changes

Raffers:
## Purpose
LPAL-250 - The user LPA list has a h1 heading for each LPA in the list. There should only be one h1 heading on a page, otherwise, it causes confusion when navigating via screen reader.

LPAL-246 - Focus skips from the right-hand button to the link which is on its left-hand side when in attorney list. This goes against the expected left-right order when navigating by keyboard (in the UK).

Fixes LPAL-250 & LPAL-246

## Approach

Makes minor HTML & CSS changes

## Learning

## Checklist


service-front/module/Application/view/application/authenticated/dashboard/macros.twig
```diff
@@ -15,13 +15,13 @@
             <li class="list-item list-item--v3 status-container--{{ status }}" {% if lpa.refreshId %}data-refresh-id="{{ lpa.refreshId }}"{% endif %}>
                 <div class="info">
                     <div class="main-info">
-                        <h1 class="list-item_title">
+                        <h3 class="list-item_title">
```
townxelliot, line 18: Quick one, @Raffers: is this list always displayed under an h2? I think it follows a h1, so should be an h2, otherwise that introduces a different accessibility issue (incorrect heading numbering).

service-front/module/Application/view/application/authenticated/lpa/primary-attorney/partials/attorney.twig
```diff
@@ -4,10 +4,10 @@
         <p class="person-address flush--ends">{{ attorney['address'] }}</p>
         <ul class="person-actions">
             <li class="action-item">
-                <a href="{{ editUrl }}" role="button" class="action-edit js-form-popup" aria-label="View/edit Attorney, {{ attorney['name'] }}">View/edit details</a>
+                <a href="{{ confirmDeleteUrl }}" class="action-delete js-form-popup">Delete</a>
```
townxelliot, line 7: This element needs to be moved after the edit button. The focus still moves incorrectly between the buttons (from right to left).

townxelliot: It might even be possible to add a cypress test for this, which I'd recommend, to prove the fix.

townxelliot (changes requested):

townxelliot:
@Raffers I was also considering whether the text "View/Edit" on the buttons makes sense, given that we use "Change" everywhere else. Just thinking it might be better if it was "View/Change" as per the design system guidance. Might be one to clear with Amy. We'll also need her to clear the styling when you change the heading level.

townxelliot:
> @Raffers I was also considering whether the text "View/Edit" on the buttons makes sense, given that we use "Change" everywhere else (we will, at least, when we merge #227). Just thinking it might be better if it was "View/Change" as per the design system guidance. Might be one to clear with Amy if we do; or potentially create a bug for that. (We'll also need her to clear the styling when you change the heading level if it affects the font size.)

@Raffers I've fixed this across the app as I realised these are links with accessibility issues. Please ignore this comment.

service-front/module/Application/view/application/authenticated/lpa/primary-attorney/partials/attorney.twig
townxelliot: Actually, I can see I'm going to have some fun merging with this branch :)

townxelliot: @Raffers This issue is also present on the attorneys page. I'd recommend searching for every page with a View/edit link and modifying all of them. (I know this is outside scope of the issue, but only because there isn't a single partial which does all of the view/edit lists, which is what I was hoping for.)

townxelliot: Also note that the float needs to stay float:left, it's actually the element order which is causing them the order on the page to be incorrect.

townxelliot: And I also noticed that the HTML appears to be wrong on this page.

To summarise, this is what I see on replacement attorneys:

```
        <ul class="person-actions">
            <li class="action-item">
                <a href="/lpa/52614158710/replacement-attorney/confirm-delete/0" class="action-delete js-form-popup">Delete</a>
            </li>
                <a href="/lpa/52614158710/replacement-attorney/edit/0" role="button" class="action-edit js-form-popup" aria-label="View/edit Replacement Attorney Mr Dead Pool">View/edit details</a>
            <li class="action-item">
            </li>
        </ul>```

with this CSS applied to `.action-item`:

```
@media (min-width: 641px)
.person-actions .action-item {
    margin-left: 8px;
    float: right;
}
```

when what we need is this:

```
        <ul class="person-actions">
            <li class="action-item">
                <a href="/lpa/52614158710/replacement-attorney/edit/0" role="button" class="action-edit js-form-popup" aria-label="View/edit Replacement Attorney Mr Dead Pool">View/edit details</a>
            </li>
            <li class="action-item">
                <a href="/lpa/52614158710/replacement-attorney/confirm-delete/0" class="action-delete js-form-popup">Delete</a>
            </li>
        </ul>```

with this CSS:

```
@media (min-width: 641px)
.person-actions .action-item {
    margin-left: 8px;
    float: left;
}
```

service-front/module/Application/view/application/authenticated/dashboard/macros.twig
Raffers: The list is always displayed under an `h2` element, just double checked.

service-front/module/Application/view/application/authenticated/lpa/primary-attorney/partials/attorney.twig
townxelliot: Just realised this is wrong: Delete is to the left of View/edit in the live site (seems wrong to me, but there you go). So ignore the above and I'll approve this PR.

townxelliot (dismissed):

townxelliot (approved):
lgtm

PR #230, opened 2021-01-20
LPAL-252 Ensure skip link correctly moves user over navigation bar

townxelliot:
## Purpose

Clicking on the skip link when logged in currently moves the user to the top of the navigation. The DAC report mentions that this is undesirable, and that the user should skip all repeated navigation elements at the top of the page.

Fixes LPAL-252 (NB the branch name has the wrong issue number on it :( )

## Approach

Add a new anchor which is at the top of the content, before the error message (if present) but after navigational elements.

## Learning

Learned more about cypress. Had to figure out a way to mimic a screen reader, which I did by adding a step which removes stylesheets to reveal the skip link.

## Checklist


johnnolan (approved):

townxelliot:
Apologies @johnnolan for dismissing your review. I noticed an issue with the pr_build where it failed due to a11y checks. Just pushed a couple of commits to hopefully fix this.

townxelliot:
> Apologies @johnnolan for dismissing your review. I noticed an issue with the pr_build where it failed due to a11y checks. Just pushed a couple of commits to hopefully fix this.

Appears to have passed, once more open for review.

williamfalconeruk (approved):

PR #231, opened 2021-01-21
Reinstate logger for admin to fix admin side

nickdavis2001:
## Purpose
_In previous PR https://github.com/ministryofjustice/opg-lpa/pull/222 the composer.lock with opg-logger removed, was pushed by service-admin by mistake instead of service-api.  This is now fixed, along with a cypress test to check and a couple of a11y test fixes along the way_

Fixes LPAL-279

## Approach

_Reinstate old composer.lock, add a Cypress test,  minor mods to Cypress a11y tests along the way_

## Checklist


Raffers (approved):

Raffers (approved):

cypress/support/index.js
```diff
@@ -29,3 +29,8 @@ Cypress.env("email","caspertests+" + userNumber + "@lpa.opg.service.justice.gov.
 Cypress.env("password", "Cypress" + userNumber)
 Cypress.env("seeded_email","seeded_test_user@digital.justice.gov.uk");
 Cypress.env("seeded_password","Pass1234");
+// see if we've inherited an adminUrl, if not default to localhost
+var adminUrl = Cypress.env("adminUrl");
+if (adminUrl === undefined) {
+    Cypress.env("adminUrl","https://localhost:7003");
```
williamfalconeruk, line 35: wondering if this should fail, rather than default?  making sure it fails if empty might be cleaner.

williamfalconeruk (commented):
other than the comment, ok to go with this.

cypress/support/index.js
nickdavis2001: Agreed. Here is a story for it https://opgtransform.atlassian.net/browse/LPAL-280

PR #254, opened 2021-01-27
LPAL-247 Help system focus shifts to improve accessibility

townxelliot:
## Purpose

Fix issues with focus shift in the help overlay.

Fixes LPAL-247

## Approach

* Modify DOM element order so that content is focused first when opening help overlay
* Modify js to switch focus back to content when user selects help topic

## Learning

The HTML for guide is stored client-side in session storage. If you modify any of the templates, you need to clear this cache, otherwise you won't see your changes.

Setting tabindex="-1" on an element forces focus onto it, even if it's an element the user can't interact with. I used this to implement shifting focus back to the content after the user clicks on a help topic in the right-hand menu.

help == guidance

## Checklist


Raffers (approved):

townxelliot:
Note this is awaiting UAT; please don't merge yet.

PR #258, opened 2021-01-28
Cypress createlpa PF and HW next few pages after donor page

nickdavis2001:
## Purpose
_LPAL-270, LPAL-271 Migration of tests from Casper to Cypress up to and including Primary Attorney page_

## Approach

- Casper tests re-implemented as feature files and step js files using cypress.
- A few files removed from git that are interim files from grunt build and have no business being checked in. Extra step added to Homepage test to sanity check that we can still request all link and script files that are listed. This also has the benefit of protecting us against rolling out deploy with broken stylesheets.
- data-cy tag done in js for name-title
- Extra data-cy tags added to twigs. There is repetition here but the plan is mark all of them up first, make the tests work, then ultimately would hope to componentise these twigs more

## Checklist


cypress.json
```diff
@@ -3,5 +3,6 @@
     "testFiles": "**/*.feature",
     "postLogoutUrl": "https://www.gov.uk/done/lasting-power-of-attorney",
     "rootRedirectUrl": "https://www.gov.uk/power-of-attorney/make-lasting-power",
-    "numberOfGuidanceHelpTopics": 22
+    "numberOfGuidanceHelpTopics": 22,
+    "defaultCommandTimeout": 20000
```
townxelliot, line 7: I'd like to remove the defaultCommandTimeout setting and leave it as is: all this seems to achieve is slowing down the time it takes for tests to fail, rather than helping them pass.

cypress/integration/CreateHwLpa.feature
```diff
@@ -74,7 +76,16 @@ Feature: Create a Health and Welfare LPA
...
             | address-postcode| PO38 1UL |
-        Then I click "form-cancel"
-        Then I click "save-and-continue"
-        And I am taken to the when lpa starts page
-        And I see "When can the LPA be used?" in the page text
+        When I click "form-cancel"
+        And I click "save-and-continue"
+        Then I am taken to the life sustaining page
+        And I see "Who does the donor want to make decisions about life-sustaining treatment?" in the page text
+        # in this test we check CanSustainLife-0 (no option) exists, then a few lines down we actually click canSustainLife-1 (yes)
```
townxelliot, line 83: Just an FYI, when I fixed the radio button focus order, the quickest thing to do was to give the first option the ID which was previously given to the enclosing fieldset. This means that the links in the error summary will move focus onto the first radio button. If I didn't do this, it would mean reworking the whole of the multiselect and error summary elements in PHP/Twig so that the error messages point to the first radio button (which doesn't have a predictable ID). The upshot is that any tests which look for radio buttons may have to use different IDs. On the other hand, I did add a data-cy attribute to every one so you don't have to find them by ID in tests. I will figure this out when I merge from master (after we've merged this PR), but just a heads-up for the future.

cypress/integration/Signup.feature
```diff
@@ -37,8 +37,8 @@ Feature: Signup
         Then I see "There is a problem" in the page text
 
         # todo - note we should be selecting Mr like we do in the Valid scenario, not typing it in here, but due to system bug, after a previous error we get a text box
-        # instead of a dropdown. The line below will therefore need to change to When I select Mr on old style id, once that bug is fixed
-        When I type "Mr" into old style id "#name-title"
+        # instead of a dropdown. The line below will therefore need to change to When I select Mr on name-title, once this bug is fixed
```
townxelliot, line 40: Is there a bug raised for this @nickdavis2001 ?

cypress/integration/common/i_am_taken_to.js
```diff
@@ -5,43 +5,60 @@ var lpaType = Cypress.config().baseUrl + '/lpa/type';
...
 })
 
 Then(`I am taken to the dashboard page`, () => {
-  cy.url().should('eq',dashboard);
+    cy.url().should('eq',dashboard);
 })
  
 Then(`I am taken to the lpa type page`, () => {
-  cy.url().should('eq',lpaType);
+    cy.url().should('eq',lpaType);
```
townxelliot, line 20: Where is lpaType set?

cypress/integration/common/i_pause.js
```diff
@@ -0,0 +1,7 @@
+import { Then } from "cypress-cucumber-preprocessor/steps";
+
+// Pause.  This is not recommended best practice but due to Cypress being too quick at doing certain things
+// we do this in (minimal possible) places until we have fixed underlying issues
+Then('I pause', () => {
```
townxelliot, line 5: Still don't like this. Is there any way to just keep checking the number of elements in the drop-down until it "settles" and has the correct list in it? Alternatively, is it possible to blat the search term into the box without typing it, so the lookahead doesn't start populating with a partial postcode? (is that what force fill does?)

townxelliot (changes requested):
Nice, just a few questions.

cypress/integration/common/i_am_taken_to.js
nickdavis2001: Top of the same file

cypress/integration/Signup.feature
nickdavis2001: Yes, did when previously encountered it

cypress.json
nickdavis2001: Fair enough. done :)

cypress/integration/common/i_pause.js
nickdavis2001: This is 100% a temp fix. Its either, leave this in for now or take it out and tests will currently fail. The intention of the wait is to make sure Cypress finished typing in the postcode, before hitting the button. I intend to find a way for Cypress to double check its finished typing. i:e re-inspect the box to see if it contains what Cypress should have typed in. Feels a bit wrong though. Cypress shouldn't be behaving like this. Am working on a long-term solution is basically the answer to this.

cypress/integration/CreateHwLpa.feature
nickdavis2001: Thanks for the info :)

cypress/integration/common/i_pause.js
townxelliot: Would it make sense to raise a bug specifically for this? Is it not possible to force the value on the element, or do you have to type for the lookahead to work?

cypress/integration/common/i_am_taken_to.js
townxelliot: Sorry, didn't notice that.

cypress/integration/common/i_pause.js
nickdavis2001: OK, I now replaced the i_pause with an explicit check that the element is filled in. This should keep trying.  (force doesn't apply to this situation I don't believe - the issue seems to simply be cypress hitting the lookup button before finishing typing)

townxelliot (approved):
lgtm, prefer this to a random pause; in this case we now pause but keep looking for something and stop pausing when we find it: far more agreeable

townxelliot (approved):
lgtm

townxelliot (approved):

PR #262, opened 2021-01-29
Dependabot php upgrades

williamfalconeruk:
## Purpose
update multiple composer.lock files to upgrade `aws-sdk-php`

- service-pdf
- service-api
- tests

Fixes LPAL-285

## Approach

`composer outdated`
`composer upgrade`

## Learning
N/A

## Checklist


townxelliot (commented):
Should we think about updating composer.json as well as the lock files? If we only update the lock files we could potentially find that the composer.json drifts away from it, such that our local environment could cease to resemble the actual environment.

(My question is probably just because of my ignorance of how to manage a composer project properly.)

williamfalconeruk:
> Should we think about updating composer.json as well as the lock files? If we only update the lock files we could potentially find that the composer.json drifts away from it, such that our local environment could cease to resemble the actual environment.
> 
> (My question is probably just because of my ignorance of how to manage a composer project properly.)

i think because some of the composer.json semver versions are a bit looser it will attempt to update the lock file, so as to bring in the minor updated versions. I think it would be worth revisiting the composer.json files to see what can be upgraded, but running composer update doesn't seem to do anything to those files.

townxelliot:
> > Should we think about updating composer.json as well as the lock files? If we only update the lock files we could potentially find that the composer.json drifts away from it, such that our local environment could cease to resemble the actual environment.
> > (My question is probably just because of my ignorance of how to manage a composer project properly.)
> 
> i think because some of the composer.json semver versions are a bit looser it will attempt to update the lock file, so as to bring in the minor updated versions. I think it would be worth revisiting the composer.json files to see what can be upgraded, but running composer update doesn't seem to do anything to those files.

OK, so it might be that composer.json has a sufficiently lax version specified that it will update just the lock file without needing an update to composer.json itself. My only concern about this is that composer.json might be too lax, and allow minor version updates which can potentially break stuff (e.g. if maintainers don't know what they're doing and break an API in a minor version). I think the functional tests cover this for now, though, but like you say probably worth keeping an eye on longer term.

townxelliot (approved):

PR #272, opened 2021-02-03
LPAL-248 Radio button focus order

townxelliot:
## Purpose

https://opgtransform.atlassian.net/browse/LPAL-248

Fixes LPAL-248

## Approach

Error summary links for radio button sets go to the ID of the fieldset around the radios using "'#" in the URL, which link direct to an ID. Rather than come up with a way to assign a new ID to the radio buttons and link to that, I moved the ID from the fieldset to the first radio button.

To implement:

* Remove the ID for the set of radio buttons from the fieldset
* Assign that ID to the first radio button instead

The FormRadio helper is where I put the logic for setting the ID. I set an ID on the FormRadio (which actually renders a set of radio buttons), then when that's rendered it's set on the first radio.

Unfortunately, some parts of the code base do radio button rendering in a completely different way, presumably because the FormRadio can't cope with HTML or large blocks of text or unusual layouts. This means that there are some bodges in this PR where I haven't replaced the existing HTML templates but have just made them work with the altered FormRadio.

I also modified the casper tests as many relied on a radio button ID to select it. Instead, I have changed these over to use a name+value selector. In the cypress tests we use a data-cy, so to make life simpler I added this to all radios rendered through FormRadio.

Apologies for the whitespace removal noise in the cypress features. I have my text editor set up to remove these on save.

## Learning

FormRadio is very nasty. Steer clear if you can.

## Checklist


nickdavis2001 (approved):
lgtm. :)

PR #286, opened 2021-02-08
LPAL-245 Session timeout focus

townxelliot:
## Purpose

When the session times out, ensure that focus remains within the session timeout popup.

https://opgtransform.atlassian.net/jira/software/projects/LPAL/boards/173?selectedIssue=LPAL-245

Fixes LPAL-245

## Approach

The focus fix was fairly trivial: I modified the JS on the client side to properly hold focus on elements within the popup. Previously, only key events on the buttons were captured; instead, I ensured that the popup was focused and captured all key events on it. It should now have to be closed before the user's focus can move outside it.

Testing this was the hard part.

Originally, I added some debug code to the client-side JS so that tests could reduce the timeout period. This had the advantage of being self-contained and only touching one file, but the team didn't like this approach. On reflection, it potentially enables a user to set their own timeout so that they are never logged out of the app, which is a potential security issue.

Instead, I made session timeouts configurable by environment variable. This has the knock-on of making the local app timeout after 30 seconds of inactivity (to enable the cypress tests to run within a reasonable time), which is fairly annoying.

In fact, it was so annoying that I returned to the client-side mechanism, after further discussion with @moretonb and @nickdavis2001. This enables appending a query string variable to simulate nearing the end of a session, as described in the [manual acceptance criteria guidance](https://opgtransform.atlassian.net/jira/software/projects/LPAL/boards/173?selectedIssue=LPAL-245) on the issue.

Given the API for fetching the session length is client-side and can be manipulated in JS anyway, I considered this to be as secure as the existing code.

However, I then had another rethink, and added some API to enable the session length to be manipulated dynamically. I then modified the tests to artificially shorten the session length during the test run so that I could force a quick session timeout popup.

## Learning

This touched more areas of the code than I'd like.

Long term: consider spinning up a local environment specifically for testing, and/or make running environment configurable at runtime (e.g. to reset the session timeout just for the duration of the test). Having to adjust the session timeout for the local stack as a whole means that it behaves differently from the live one, and also times sessions out after 30 seconds on a page.

Another approach might be to disable session timeouts in the local stack, perhaps.

## Checklist


terraform/environment/ecs_api.tf
```diff
@@ -297,6 +297,7 @@ locals {
         { "name" : "OPG_DOCKER_TAG", "value" : var.container_version },
         { "name" : "OPG_LPA_STACK_ENVIRONMENT", "value" : local.account_name },
         { "name" : "OPG_LPA_COMMON_APPLICATION_LOG_PATH", "value" : "/var/log/app/application.log" },
+        { "name" : "OPG_LPA_AUTH_TOKEN_TTL", "value" : tostring(local.account.auth_token_ttl_secs) },
```
williamfalconeruk, line 300: 👍🏻  looks good!

terraform/environment/terraform.tfvars.json
```diff
@@ -16,6 +16,7 @@
       "is_production": "false",
       "sirius_api_gateway_endpoint": "https://api.dev.sirius.opg.digital/v1/lpa-online-tool/lpas/",
       "sirius_api_gateway_arn": "arn:aws:execute-api:eu-west-1:288342028542:*/*/GET/lpa-online-tool/*",
+      "auth_token_ttl_secs": 330,
```
williamfalconeruk, line 19: thinking - is 330 seconds too long?

townxelliot: From the perspective of the tests, they have to wait 35 seconds before the session timeout popup appears. It seems OK to me. A bigger concern is whether this is too short when you're developing, as you get timed out (locally) every 30 seconds.

docker-compose.yml
```diff
@@ -189,6 +189,9 @@ services:
 
       OPG_LPA_PROCESSING_STATUS_ENDPOINT: "http://gateway:5000/v1/lpa-online-tool/lpas/"
 
+      # lifetime (in seconds) of authentication tokens; in prod this defaults
+      # to 4500, but we set it to 5.5 minutes here so we can test session timeouts
+      OPG_LPA_AUTH_TOKEN_TTL: 330
```
townxelliot, line 194: change this to 4500

cypress/integration/Session.feature
```diff
@@ -0,0 +1,34 @@
+@Session
+Feature: Session
+
+    I want to be notified when my session is about to expire
+
+    Background:
+        Given I ignore application exceptions
+
+    @focus
+    Scenario: Session timeout pop-up displays after inactivity and keeps focus (LPAL-245)
```
townxelliot, line 10: tag this test so it only runs in CI

townxelliot (commented):
After discussion with @nickdavis2001 and Moreton, I've decided to reinstate the query string variable to enable shortened sessions in dev/test (see https://github.com/ministryofjustice/opg-lpa/pull/286/commits/1f17be8500f12df25ca352922a30d2a53452f1c7). This is because a short session in dev and UAT will cause problems where we are working on a page for a long period; also because the current arrangement means that UAT and prod have different configurations which may conceal differences in behaviour.

service-front/assets/js/opg/session-timeout-dialog.js
```diff
@@ -116,24 +116,40 @@ var SessionTimeoutDialog = function (options) {
...
-                logoutButton.focus();
+    this.trapNavigation = function (focusables) {
+        var currentElementIndex = 0;
+        focusables[currentElementIndex].focus();
+
+        var lastFocusableIndex = focusables.length - 1;
+
+        this.element.keydown(function (e) {
+            // capture only tab events on this dialog
+            if (e.key !== 'Tab') {
```
johnnolan, line 127: I think there may be some browser compatibility issues if you don't include the keycode option too `(e.key === 'Tab' || e.keyCode === 9);`

service-front/assets/js/opg/session-timeout-dialog.js
```diff
@@ -116,24 +116,40 @@ var SessionTimeoutDialog = function (options) {
...
-            if (e.key === 'Tab' && e.shiftKey) {
-                e.preventDefault();
-                logoutButton.focus();
+    this.trapNavigation = function (focusables) {
+        var currentElementIndex = 0;
+        focusables[currentElementIndex].focus();
+
+        var lastFocusableIndex = focusables.length - 1;
+
+        this.element.keydown(function (e) {
```
johnnolan, line 125: It may be out of scope, but a common practice is to close the dialog by using the esc key, can see this here as an example if it fits here https://github.com/ministryofjustice/opg-use-an-lpa/blob/6a0b37c7f4964740b381266f611df03f80b1e951/service-front/web/src/javascript/sessionDialog.js#L92

townxelliot: It's not out of scope, I'm just not sure how that key would be interpreted here: we have "OK" and "Sign out" as options, while "Esc" doesn't really signify commitment to either of those, so I'm not sure what it would do. That's what I didn't add a handler for it.

service-front/assets/js/opg/session-timeout-dialog.js
townxelliot: Thanks @johnnolan, fixed in 8ab4fc7e4

service-front/assets/js/opg/session-timeout-dialog.js
johnnolan: Makes sense, in that case maybe its one to discuss separately where we can have a think about the user journey?

townxelliot: I suggest raising a separate issue for that as it's not within the scope of this one. I'll do that.

cypress/integration/Session.feature
townxelliot: After modifying how this works, it's now possible to modify the session timeout via a query string variable, so this tag is no longer necessary.

service-front/assets/js/opg/session-timeout-dialog.js
townxelliot: @johnnolan I raised https://opgtransform.atlassian.net/browse/LPAL-307 for this if you'd like to expand/refine it.

johnnolan (approved):

nickdavis2001 (dismissed):

townxelliot:
@williamfalconeruk I have reworked this so that we now have an (authenticated) endpoint in service-api which can set the session expiry time to any number of seconds. The tests now invoke this API to set the session to 302 seconds, so that the session timeout dialog displays after a 3 second wait during the tests. This enables us to remove the session timeout code in the test which was based on JS and is a better solution. Unfortunately, this touches a lot of code, so I'll need another review please.

service-api/module/Application/src/Model/Service/ProcessingStatus/Service.php
```diff
@@ -101,12 +101,12 @@ public function getStatuses($ids)
             ],
             'fulfilled' => function ($response, $id) use (&$results) {
                 // Each successful response
-                $this->getLogger()->debug('We have a result for:' . $id);
+                //$this->getLogger()->debug('We have a result for:' . $id);
```
williamfalconeruk, line 104: not sure if this was right, but debug logging commented out here and below.

townxelliot: good spot, that was to reduce noise while I was working on this, I'll reinstate it

service-front/assets/js/opg/session-timeout-dialog.js
```diff
@@ -108,32 +132,48 @@ var SessionTimeoutDialog = function (options) {
...
-        logoutButton.keydown(function (e) {
-            if (e.key === 'Tab' && !e.shiftKey) {
-                e.preventDefault();
-                continueButton.focus();
+            e.preventDefault();
+
+            var direction = 1;
+            if (e.shiftKey) {
+                direction = -1;
             }
```
williamfalconeruk, line 159:
```suggestion
            var direction = (e.shiftKey ? -1 : 1); 
```
might be more succinct, though is personal choice...

townxelliot: I think I avoided this in case it's not supported by older browsers; but as this file is compiled by the minifier anyway, that's maybe not a consideration.

williamfalconeruk: it is probably  minified this way anyway, like I said just a personal preference of mine so no issues 👍🏻

cypress/integration/common/elements.js
```diff
@@ -12,17 +12,39 @@ const MapNumberTextToNumber = {
 // map from natural language text to HTML tag name
 const MapElementSpecifierToTag = {
     'level 1 heading': 'H1',
-    'level 2 heading': 'H2'
+    'level 2 heading': 'H2',
```
williamfalconeruk, line 15: not sure if this was intentional - but there's an additional comma here. i know PHP allows this...

service-front/module/Application/src/Controller/Authenticated/SessionKeepAliveController.php
```diff
@@ -12,4 +12,34 @@ public function indexAction()
...
+                return $response;
+            }
+
+            $remainingSeconds = $this->getAuthenticationService()->setSessionExpiry(intval($expireInSeconds));
+            return new JsonModel(['remainingSeconds' => $remainingSeconds]);
+        }
+
+        $response = $this->getResponse();
+        $response->setStatusCode(405);
+        $response->setContent('Bad request method');
```
williamfalconeruk, line 42: 405 should be `method not allowed` rather than bad request method (which is more like the 400 above) - which is more relevant to this type of api error. (might be being pedantic!)

williamfalconeruk (approved):
🏆

PR #288, opened 2021-02-08
LPAL-270 and LPAL-271 Create LPA casper to cypress

nickdavis2001:
## Purpose
_LPAL-270 and LPAL-271 moving Casper tests to Cypress, for attorney decisions and certificate provider_

## Approach

- Re-implemented casper tests in cypress as feature files and step js implementation.
- Add data-cy tags to relevant twig files.
- Corrected a couple of console log lines in casper tests which mentioned primary when they should have said replacement.
- Added step to test whether an element is found within the error wrapper. Used this for the casper test that I was replacing, also used it in Feedback initally as a POC. In general, since we now have this step to do it for us, it would be nice eventually to check that the correct fields get wrapped with the red bar, not merely check for error text.

## Learning

## Checklist


cypress/integration/common/i_click.js
```diff
@@ -20,6 +20,21 @@ Then(`I click third occurrence of {string}`, (clickable) => {
     cy.OPGCheckA11y();
 })
 
+Then(`I click fourth occurrence of {string}`, (clickable) => {
```
townxelliot, line 23: Could you reduce the repeated code here by creating a "I click {nth} occurrence of {string}"?

townxelliot (changes requested):

cypress/integration/common/i_click.js
nickdavis2001: Started out trying to make it nice English. Then there turned out to be lots of occurrences.! Have now changed this to "I click occurrence {int} of {string}" . Note that this counts up from 0. Toyed with the idea of doing -1 and having it start at 1 in the feature file, but thought that would be confusing

townxelliot (dismissed):

townxelliot (commented):
@nickdavis2001 you could leave this in the Errors.feature file and just replicate the Background steps in that file for that single test.

townxelliot (approved):

PR #293, opened 2021-02-10
LPAL-255 Fix email text overflow in help/guidance popup

townxelliot:
## Purpose

Fix accessibility issue where email address on the guidance popup (at the bottom of the popup) overflowed the text box.

Fixes LPAL-255

## Approach

Allow words to break within the guidance popup. This causes the email address to wrap within the text box.

## Learning

It's not possible to test browser zoom in JS. Instead, I made the viewport really narrow to force the text to wrap. I then applied the technique of comparing scrollWidth to clientWidth, to ensure that the clientWidth was at least as big as the scrollWidth (i.e. the text in the block element doesn't need scrollbars, therefore it is either enclosed in the textbox or wrapping).

## Checklist


service-front/assets/sass/patterns/_help-system.scss
```diff
@@ -185,6 +185,12 @@
 .help-footer {
     padding: $gutter 0 0;
 
+    p {
+        overflow: hidden;
```
williamfalconeruk, line 189: nice bit of CSS - how i miss messing with this stuff 😁

williamfalconeruk (commented):
Looks ok to me!

williamfalconeruk (approved):
Approved

service-front/assets/sass/patterns/_help-system.scss
townxelliot: I could probably just add break word and not the other styles, but these don't hurt as they only affect this one paragraph and help it display better where the word breaks aren't supported (e.g. very old browsers).

williamfalconeruk: makes sense

townxelliot:
I'll release this tomorrow morning and update from master then, too.

PR #317, opened 2021-02-24
LPAL-309 - Checkbox Fieldset associated error

Raffers:
## Purpose
Associate error message with the Correspondence Fieldset when no options have been selected

Fixes LPAL-309

## Approach

Updates twig template to display the correct error, so it appears to be associated with the elements of the fieldset

## Learning

## Checklist


service-front/module/Application/view/application/authenticated/lpa/correspondent/index.twig
```diff
@@ -136,13 +136,16 @@
...
 
-            <fieldset id="correspondence">
-
+            <fieldset id="{{ correspondence.getAttribute('name') }}">
                 <legend class="visually-hidden">Contact options</legend>
                 <h3 class="heading-medium flush--top">How would the person above prefer to be contacted?</h3>
 
                 <p>Select how the correspondent would like to be contacted. You can choose more than one option.</p>
 
+                {% if topLevelError %}
```
Raffers, line 145: This is required as if there is an error on either/both of phone/email elements they will also display the error message here. Which is not required & would cause confusion.

service-front/module/Application/src/Form/Fieldset/Correspondence.php
```diff
@@ -1,23 +1,23 @@
 <?php
 
-namespace Application\Form\Lpa;
+namespace Application\Form\Fieldset;
 
 use Laminas\Form\Fieldset;
 use Laminas\Hydrator\ClassMethodsHydrator as ClassMethodsHydrator;
-use Opg\Lpa\DataModel\Lpa\Document\Correspondence;
+use Opg\Lpa\DataModel\Lpa\Document\Correspondence AS CorrespondenceModel;
```
townxelliot, line 7: Suggest lowercase "AS"

townxelliot (changes requested):
@Raffers is this the only place where the error messages are displayed with a checkbox, or does this fix the issue for all checkboxes and their error messages?

service-front/module/Application/view/application/authenticated/lpa/correspondent/index.twig
townxelliot: Have to be careful here: the ID should be on the top checkbox in the set, rather than on the fieldset. This is so that the message shown in the error summary at the top of the page links to the checkbox, not the surrounding fieldset; see https://opgtransform.atlassian.net/browse/LPAL-248

townxelliot (changes requested):

service-front/module/Application/view/application/authenticated/lpa/correspondent/index.twig
```diff
@@ -136,13 +136,16 @@
 
         <div class="form-group {{ topLevelError ? 'form-group-error'}}">
 
-            <fieldset id="correspondence">
-
+            <fieldset id="{{ correspondence.getAttribute('name') }}">
```
townxelliot, line 139: Have to be careful here: the ID should be on the top checkbox in the set, rather than on the fieldset. This is so that the message shown in the error summary at the top of the page links to the checkbox, not the surrounding fieldset; see https://opgtransform.atlassian.net/browse/LPAL-248

townxelliot (changes requested):

townxelliot:
@Raffers is there any way we can do integration tests for this, e.g. create an error on the page, click on the error link, ensure focus is in the top checkbox, ensure error message is above the set of checkboxes?

service-front/module/Application/view/application/authenticated/lpa/correspondent/index.twig
Raffers: This was to bring it inline with the other fieldsets, that I came across. They were all setup in this fashion. There will only be one error on here for checkboxes & that is to make sure at least one has been selected. The ID is still the same value, it's now defined `CorrespondenceForm` instead.

Raffers:
> @Raffers is there any way we can do integration tests for this, e.g. create an error on the page, click on the error link, ensure focus is in the top checkbox, ensure error message is above the set of checkboxes?

I could roll it into the current LPA tests, but as we currently have 3 sets of these I wanted to avoid bloating these more unnecessarily & thought it would be better to wait. Until we have proper mocking in place for each stage of the LPA

service-front/module/Application/src/Form/Fieldset/Correspondence.php
Raffers: Hangup from doing so much sql 😅

townxelliot (approved):

PR #327, opened 2021-03-01
LPAL-311 - Re-order the display of error messages, to above the input element, but below any labels.

Raffers:
## Purpose
Errors should be placed above the field where they occurred; currently, they’re in a separate box or under the field.

Fixes LPAL-311

## Approach

Adjusts HTML macros & other twig templates to correctly place the error element, adjusts CSS to display correctly
Also cleans up some of the twig templates to make them more readable/condensed.

## Learning

## Checklist


townxelliot (changes requested):
Positioning is slightly off for correspondent error messages:
Screenshot at 2021-03-05 11-04-07

townxelliot (changes requested):

townxelliot:
@Raffers the layout in the popups when displaying error messages is off (I'm on Chrome), but good apart from that, thanks.

townxelliot:
@Raffers the padding on date error messages can (I think) be adjusted independently of the padding setting for other errors, e.g.

Screenshot at 2021-03-12 17-04-39

The CSS looks like this:

```
@media (min-width: 641px)
.form-popup .form-date .error-message {
    top: 3px;
    left: 0;
    width: 100%;
    padding-bottom: 0;
}
```

Note the `padding-bottom: 0`, which overrides the padding only on error messages inside a form-date. You would need to figure out where this is in the SASS files, but should be possible to selectively fix the padding this way. Happy to have a look with you on Monday if that would help.

townxelliot (changes requested):
Good stuff, but I still think we can do better positioning of the error message above date boxes; and I'm not sure about the alignment of the error messages with the labels for input fields. We might want to think about whether it's worth taking the time to make that better, too.

docker-compose.yml
```diff
@@ -98,7 +98,7 @@ services:
 
       PHP_OPCACHE_VALIDATE_TIMESTAMPS: 1
       PHP_IDE_CONFIG: serverName=lpa-front-app
-      # ENABLE_XDEBUG: 'true'
+      ENABLE_XDEBUG: 'true'
```
nickdavis2001, line 101: Did you definitely intend to check in this change?  If I recall correctly, while it'd be convenient and logical to have xdebug always turned on in dev, it was commented out because it can be horribly slow to start up, so we only want to use it when we have to debug. Not an ideal situation cos then one has to keep turning it back on. I wonder if we could use an environment variable for it actually, that would sort the issue of sometimes needing it and sometimes not?

Raffers: Nope, I keep forgetting about that.

townxelliot (approved):

PR #332, opened 2021-03-04
Python api test fixtures and stitched feature files

nickdavis2001:
## Purpose
_Call the back-end API directly from python, so that this could be called by (1) cypress tests for making test fixtures quickly, (2) dedicated api tests (3) a new front-end written in python ;)_

## Approach

- implemented python api, currently with json hard-coded as used by test fixtures, but in future this can be built upon to take arbitrary json.
- fixtures.js calls the python api to make test fixtures. (and tear down at end)
- Original huge CreateHWLpa.feature and CreatePFLpa.feature now have 3 features files stripped off the top, these are LpaType**.feature, Donor**.feature, Attorneys**.feature.  These can run independently or be stitched
- stitch.sh stitches the feature files back together for running from command line or in CI

_

## Checklist


service-api/docker/web/etc/confd/templates/nginx.conf.dev
```diff
@@ -0,0 +1,115 @@
+log_format tracing escape=json '{'
```
townxelliot, line 1: Whereabouts is this used in the build?

tests/python-api-client/testapi.py
```diff
@@ -0,0 +1,20 @@
+from lpaapi import *
```
townxelliot, line 1: As a next step, would be nice to add some GETs to check that the API has added what you expect (out of scope for this issue but could raise separately).

townxelliot (commented):
Just running the tests locally, looking good. Only have a couple of questions, everything else in the code looks fine.

service-api/docker/web/etc/confd/templates/nginx.conf.dev
nickdavis2001: This is used to get the service-api to log what json was sent from the front-end to the api.  It can be copied over to nginx.conf to achieve this.  I debated whether to have this file checked into git or not , as how to do this is also described in logging.md. However, since I have the file locally, seems a shame not to check it in as an example.  What's your opinion?  I was somewhat undecided....

tests/python-api-client/testapi.py
nickdavis2001: Sounds a good idea :)

service-api/docker/web/etc/confd/templates/nginx.conf.dev
townxelliot: It's OK, but worth a comment explaining why it's here and pointing to logging.md?

nickdavis2001: Agreed. Now done :)

townxelliot (approved):

PR #364, opened 2021-03-17
LPAL-316 Fix page heading levels to improve accessibility

townxelliot:
## Purpose

https://opgtransform.atlassian.net/browse/LPAL-316

Fixes LPAL-316

## Approach

Shift around the headings so that when in "accordion" mode (i.e. creating an LPA), the instructions at the top of the page are collapsed into a details element and the active part of the accordion is an h1.

## Learning

n/a

## Checklist


nickdavis2001 (approved):
lgtm

service-front/module/Application/view/application/authenticated/lpa/reuse-details/index.twig
```diff
@@ -1,4 +1,4 @@
-{% extends isPopup ? 'layout/twig/popup.twig' : 'layout/twig/layout.twig' %}
+{% extends isPopup ? 'layout/popup.twig' : 'layout/twig/layout.twig' %}
```
williamfalconeruk, line 1: noticed that the `layout.twig` here is in a different place to one i'm marking below - is this right?

service-front/module/Application/view/application/authenticated/lpa/status/index.twig
```diff
@@ -1,4 +1,4 @@
-{% extends 'layout/twig/layout.twig' %}
+{% extends 'layout/layout.twig' %}
```
williamfalconeruk, line 1: see comparison to other comment

service-front/module/Application/view/application/authenticated/lpa/reuse-details/index.twig
williamfalconeruk: there are several of these so just checking

williamfalconeruk (changes requested):
holding back just in case the twig structure is not quite lined up.

service-front/module/Application/view/application/authenticated/lpa/status/index.twig
townxelliot: I shifted all of the layout twig templates out of the folder called "twig", which was just pointless and another thing to click through when looking for files. I moved those templates up a level to the parent of "twig" and got rid of "twig". Not strictly necessary but annoys the hell out of me when working on these templates :)

williamfalconeruk: indeed - thats all good, just that the `twig` folder is still referenced in a number of places above.

townxelliot: I think I caught all the references to `layout/twig` but please shout if you can see one I've missed.

service-front/module/Application/view/application/authenticated/lpa/reuse-details/index.twig
townxelliot: Thanks for spotting these @williamfalconeruk I've now amended. Interesting that UAT still passed and the functional tests... Though this may only affect layout in very particular circumstances which aren't picked up by either.

williamfalconeruk (dismissed):
🏆

williamfalconeruk (approved):

PR #375, opened 2021-03-22
LPAL-257 Remove redundant variable and make page headings + titles consistent

townxelliot:
htmlTitle was just pageTitle with a hyphen after it: lots of
wasted characters.

Remove this and rework how the page title is displayed, so that
the content of the h1 heading and <title> element match.

For the homepage, don't put the pageTitle into the <title>
element; this is in line with the example set by
https://www.gov.uk/apply-renew-passport

## Purpose

https://opgtransform.atlassian.net/browse/LPAL-257

Fixes LPAL-257

## Approach

See above

## Learning

n/a

## Checklist


nickdavis2001 (approved):
This all LGTM.  The only thing that bothers me in relation to this, how strong are our tests?  This is a simple change but covers a lot of the site. Should some extra lines be added to cypress tests to sanity check such major changes in general? So far the tests often just check for some text existing somewhere on the page, I wonder if we ought to start tightening up a bit.  Doesn't have to be in this PR though.  So will approve anyway :)

townxelliot:
> This all LGTM. The only thing that bothers me in relation to this, how strong are our tests? This is a simple change but covers a lot of the site. Should some extra lines be added to cypress tests to sanity check such major changes in general? So far the tests often just check for some text existing somewhere on the page, I wonder if we ought to start tightening up a bit. Doesn't have to be in this PR though. So will approve anyway :)

Thanks @nickdavis2001. I did consider adding a test which checks that the content of the h1 on the page is at the start of the content in the <title> element, but wondered whether this was too much and would just slow everything down even more. I think this should be relatively simple to add as an after step, perhaps?

townxelliot:
> > This all LGTM. The only thing that bothers me in relation to this, how strong are our tests? This is a simple change but covers a lot of the site. Should some extra lines be added to cypress tests to sanity check such major changes in general? So far the tests often just check for some text existing somewhere on the page, I wonder if we ought to start tightening up a bit. Doesn't have to be in this PR though. So will approve anyway :)
> 
> Thanks @nickdavis2001. I did consider adding a test which checks that the content of the h1 on the page is at the start of the content in the <title> element, but wondered whether this was too much and would just slow everything down even more. I think this should be relatively simple to add as an after step, perhaps?

I added the test suggested and amended a few titles I'd missed which only appear under specific conditions, as identified in UAT.

nickdavis2001 (approved):
lgtm :)

PR #387, opened 2021-03-23
Npm cypress make runner added

williamfalconeruk:
## Purpose

to add the option to run `cypress open` via the makefile

Dev improvement.

## Approach

- add documentation
- add additional package files
- add makefile entry

## Learning

https://docs.cypress.io/guides/getting-started/installing-cypress

## Checklist


Makefile
```diff
@@ -176,6 +176,10 @@ cypress-local:
 cypress-local-shell:
 	aws-vault exec moj-lpa-dev -- docker run -it -e AWS_ACCESS_KEY_ID -e AWS_SECRET_ACCESS_KEY -e AWS_SESSION_TOKEN -e "CYPRESS_baseUrl=https://localhost:7002" --entrypoint bash --network="host" -v `pwd`/cypress:/app/cypress cypress:latest
 
+.PHONY: cypress-gui-npm
```
nickdavis2001, line 179: Devil's advocate Q - does putting this in the Makefile really add any value?  Cos make cypress-gui-npm is no less to type than cypress open, but the latter makes you think about what you're doing more, and works anywhere that cypress is installed.... ;)

package.json
```diff
@@ -1,5 +1,495 @@
 {
```
nickdavis2001, line 1: I don't know a lot about package.json in general but 2 thoughts - (a) this looks a bit huge, and possibly maintenance issue - what happens when we get later version. of cypress? I guess just update this file (b) can we use package.json to install any other things while we're at it. Are there any other npm packages it would be have installed.

Makefile
williamfalconeruk: it is a good point - i almost didn't put this in. more for consistency.

package.json
williamfalconeruk: a) this lot was auto installed when doing the `npm i --saveDev` for the dependancies. so should be ok, package manager will handle this. 
b) absolutely! but as and when we need it.

Makefile
williamfalconeruk: the original Readme i did listed `cypress open` as the way to run it, but then thought might as well add it. it's neither here nor there as you say in terms of keystrokes. the make does specify the folder to run, which means it auto picks up the cypress folder correctly.

README.md
```diff
@@ -121,7 +122,28 @@ run the tests:
...
+### Run the Cypress GUI via npm
+
+the recommended way to run tests for the GUI runner is using npm. see: <https://docs.cypress.io/guides/getting-started/installing-cypress/#System-requirements>. this is useful if you want to see what the tests are physically doing.
+
+Install cypress globally:
+
+`npm i -g cypress`
+
+The package.json in the root of the repo has all of the required dev dependancies for Cypress. Add plugins to this as needed using `npm i <package-name> --saveDev`
+
```
williamfalconeruk, line 134:
```suggestion
### Run the Cypress GUI via npm

the recommended way to run tests for the GUI runner is using npm. see: <https://docs.cypress.io/guides/getting-started/installing-cypress/#System-requirements>. this is useful if you want to see what the tests are physically doing.

Install cypress globally:

`npm i -g cypress`

The package.json in the root of the repo has all of the required dev dependancies for Cypress. Add plugins to this as needed using `npm i <package-name> --saveDev`

you have 2 options you can use:

For general use  and for exploratory work you can just run:

``` bash
cypress open
```

Alternatively, the makefile has a default command to point at the existing cypress tests, and will list them.

```

Makefile
williamfalconeruk: bit of rejigging done, good for a re-review also used MD lint locally and fixed a few things.

nickdavis2001 (approved):
lgtm

PR #392, opened 2021-03-24
LPAL-232 Title lost dropdown

townxelliot:
## Purpose

https://opgtransform.atlassian.net/browse/LPAL-232

Fixes LPAL-232

## Approach

Debugged how options were added to the title select element; noticed that when the JS console showed an analytics error, the options didn't render; put a guard into the JS function so that it wouldn't call a non-existent method, which allowed the JS to run properly.

## Learning

n/a

## Checklist


Makefile
```diff
@@ -199,3 +199,10 @@ endif
 .PHONY: restitch
 restitch:
 	cypress/stitch.sh
+
+# Start S3 Monitor and call "cypress open";
+# this requires a globally-installed cypress
+.PHONY: cypress-open
+cypress-open:
```
williamfalconeruk, line 206: noice 👍🏻

williamfalconeruk (dismissed):

Makefile
```diff
@@ -199,3 +199,10 @@ endif
...
+	aws-vault exec moj-lpa-dev -- python3 cypress/S3Monitor.py &
+	CYPRESS_userNumber=`node cypress/userNumber.js` cypress open
```
nickdavis2001, line 208: Nice :)

cypress/userNumber.js
```diff
@@ -1,11 +1,7 @@
...
-var date = new Date();
-var userNumber = date.getTime() + "" + random;
-console.log(userNumber);
+// This makes a user number for tests.
+// It needs to happen before startup rather than in support/index.js so
+// that we have the same user number for all feature files under CI.
+// This code is therefore called from start.sh.
+// For safety, for now we generate userNumber using
+// the exact same js code that the Casper tests did.
+console.log("" + (new Date()).getTime() + Math.floor(Math.random() * 999999999));
```
nickdavis2001, line 7: Do ya really think that's more readable than what was there before?? ;)

townxelliot: No, I think I'll revert it

nickdavis2001 (approved):
lgtm

PR #403, opened 2021-03-31
LPAL-286 PHP 8 upgrade prep

townxelliot:
## Purpose

Make various updates to service-front in preparation for moving to PHP 8. These are mostly in the nature of fixing method signatures, ensuring methods are called with the right arguments and number of arguments, and removing use of functions deprecated in PHP 8. All of this work is compatible with PHP 7.4.

Prep for LPAL-286

## Approach

see above

## Learning

composer is a pain

## Checklist


service-front/module/Application/src/Controller/Authenticated/Lpa/CheckoutController.php
```diff
@@ -36,10 +38,10 @@ public function indexAction()
         $isRepeatApplication = ($lpa->repeatCaseNumber != null);
 
         $lowIncomeFee = Calculator::getLowIncomeFee($isRepeatApplication);
-        $lowIncomeFee = (floor($lowIncomeFee) == $lowIncomeFee ? $lowIncomeFee : money_format('%i', $lowIncomeFee));
+        $lowIncomeFee = number_format(floatval($lowIncomeFee), 2);
```
townxelliot, line 41: Compare with what happens in live.

service-front/module/Application/src/Controller/Authenticated/Lpa/ReuseDetailsController.php
```diff
@@ -25,9 +25,16 @@ public function indexAction()
...
-        $callingUrl = $queryParams['calling-url'];
-        $includeTrusts = $queryParams['include-trusts'];
-        $actorName = $queryParams['actor-name'];
+
+        $callingUrl = null;
+        $includeTrusts = null;
+        $actorName = null;
+
+        if (!is_null($queryParams)) {
+            $callingUrl = $queryParams['calling-url'];
```
townxelliot, line 34: array_key_exists() here?

service-front/module/Application/src/Controller/Authenticated/Lpa/StatusController.php
```diff
@@ -45,9 +45,9 @@ public function indexAction()
         // Return either the applicationRejectedDate or the applicationRegistrationDate based on what's received from Sirius
         $metadata = $lpa->getMetadata();
 
-        if ($metadata['application-rejected-date']!= null)
+        if (array_key_exists('application-rejected-date', $metadata) && $metadata['application-rejected-date'] != null)
```
townxelliot, line 48: Array lookup method which can deal with keys not existing.

service-front/module/Application/src/Controller/Authenticated/Lpa/CheckoutController.php
townxelliot: Having looked at this and thought about it, the original code will format the cost with decimal point if it has one; otherwise it shows the cost without a decimal point, e.g.

```
cost == 42, display £42
cost == 30.50, display £30.50
```

That's why the calls to `floor()` are there, to see whether the cost has a pence component.

service-front/module/Application/src/Controller/Authenticated/Lpa/StatusController.php
townxelliot: isset() behaves in the way we require: returns false if the key is not set *or* if it is set but is null.

service-front/module/Application/src/Controller/Authenticated/Lpa/ReuseDetailsController.php
```diff
@@ -25,9 +25,16 @@ public function indexAction()
 
         //  Check that the required query params have been provided
         $queryParams = $this->params()->fromQuery();
```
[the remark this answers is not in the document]
townxelliot: I used the `??` operator for this.

nickdavis2001 (approved):
lgtm 🏆  with caveat that major changes so UAT need to be thorough

williamfalconeruk (approved):

PR #406, opened 2021-04-01
LPAL-267 Prevent ini_set errors from /ping/json endpoint

townxelliot:
## Purpose

Errors were occurring due to some session handling nonsense
going on in the AbstractControllerFactory.

By extending the AbstractActionController from Laminas directly
to give us the PingController (we don't need the other machinery
which comes with the AbstractBaseController after all), we can
remove a lot of complexity and create a controller which doesn't
need sessions.

This gets rid of the ini_set exceptions from /ping/json because
we no longer have session stuff going on.

Fixes LPAL-267

## Approach

Reduce complexity of PingController, which doesn't need session machinery.

## Learning

n/a

## Checklist

### DIFF
--- a/service-front/module/Application/config/module.config.php
+++ b/service-front/module/Application/config/module.config.php
@@ -16,6 +16,7 @@ return [
     'controllers' => [
         'factories' => [
             'DynamoDbController' => 'Application\Controller\Console\DynamoDbControllerFactory',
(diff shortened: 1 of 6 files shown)


nickdavis2001 (approved):
lgtm

williamfalconeruk (approved):

PR #417, opened 2021-04-15
LPAL-210 Cancel buttons on your details pages

townxelliot:
## Purpose

https://opgtransform.atlassian.net/browse/LPAL-210

Fixes LPAL-210

## Approach

Add some buttons.

## Learning

Don't make things more complicated than they need to be.

## Checklist

### DIFF
--- a/cypress/integration/Signup.feature
+++ b/cypress/integration/Signup.feature
@@ -3,6 +3,8 @@ Feature: Signup
 
     I want to be able to sign up
 
+    # NB these tests are order-sensitive, as some rely on the signup state
+    # being established before they run
     Background:
         Given I ignore application exceptions
 
@@ -20,6 +22,11 @@ Feature: Signup
         And I see "Account activated" in the title
         And I visit link containing "sign in"
         Then I am taken to the login page
+
+    @focus
+    Scenario: Cancel button is not shown on "Your details" the first time the user logs in (LPAL-210)
+        Given I log in as standard test user
+        Then I do not see "Cancel" in the page text
 
     @focus
     Scenario: About Me Details have Blank title, month and wrong postcode in address + long names, followed by DOB in future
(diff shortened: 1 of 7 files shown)


cypress/integration/Signup.feature
```diff
@@ -3,6 +3,8 @@ Feature: Signup
 
     I want to be able to sign up
 
+    # NB these tests are order-sensitive, as some rely on the signup state
```
williamfalconeruk, line 6: just a thought - this comment suggests there needs to be a `background` section to do this. I guess not practical yet, but does need some thought.

cypress/integration/Signup.feature
```diff
@@ -70,3 +77,29 @@ Feature: Signup
         # cypress seems t choke on being redirected off our site though
         # When I logout
         #Then I am taken to the post logout url
+
+    @focus
+    # this scenario must run after the user has saved their details for the first time
```
williamfalconeruk, line 82: just a thought - this comment suggests there needs to be a `background` section to do this. I guess not practical yet, but does need some thought.

williamfalconeruk (approved):

williamfalconeruk:
just some comments, but looks ok otherwise :)

cypress/integration/Signup.feature
townxelliot: Yes, that would be good. The whole of the "Valid user details" test needs to run for this later test to work. Another option would be to (for now) move it into its own feature which runs after signup (with the other post-signup features). At the moment, we don't have a way for signup to be a background, we are relying on test order.

townxelliot: @nickdavis2001 could we think about stitching signup into the front of tests, and having a way to put the back-end into a state where the user has signed up, instead of relying on test order?

nickdavis2001: My gut feel (though willing to be persuaded otherwise :) ) is that stitching is a bit of a hack to make things work with both Signed-up user or standard seeded user. Feels like a background step might be better here. Agreed its not ideal that these scenarios are order sensitive, and some refactoring here wouldn't hurt, long term.

townxelliot: @nickdavis2001 Shall I raise an issue to investigate this rather than it block this PR?

nickdavis2001: Background step applies to all scenarios.  I'd be inclined to do this in another PR and not block this one.

PR #426, opened 2021-04-21
Lpal 352 attempt3 of slack orb

williamfalconeruk:
## Purpose

use dynamic templates instead to generate slack message.

Fixes LPAL-352

## Approach

create new slack message template shell files which will be used to create env vars, that the orb can then parse.

## Learning

- Dynamic Templates: https://github.com/CircleCI-Public/slack-orb/wiki/Dynamic-Templates

## Checklist

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -971,58 +971,14 @@ jobs:
             scripts/pipeline/set_environment_variables/set_slack_env_vars.sh >> $BASH_ENV
             echo environment_pipeline_tasks_config.json:
             cat /tmp/environment_pipeline_tasks_config.json
+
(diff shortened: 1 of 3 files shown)


townxelliot:
Is there a reason you can't use a JSON template file and have to have it inline? Just reducing the number of files?

townxelliot (approved):
lgtm

williamfalconeruk:
> @williamfalconeruk Is there a reason you can't use a JSON template file and have to have it inline? Just reducing the number of files by inlining?

The orb seems to over-sanitize the input before using `jq` if we inline using the `custom` option. i'm not sure if my new version will work either on master yet. I can inline into the config yaml as part of the run step; it's more that this file is already large as it is, and unwieldy. We'd have to look at whether we can break it up a bit.

PR #429, opened 2021-04-21
Lpal 352 sanitize commit message for new slack orb

williamfalconeruk:
## Purpose

- Variables need substitution if they contain multiline strings, and replaced with the literal `\n`
- On top of this, we also needed to properly escape double quotes as this was messing up the orb.

Fixes LPAL-352

## Approach

- add substitution to `${COMMIT_MESSAGE}` to replace actual end of lines with  a `\n` literal. This is handled better by JQ and the new slack orb.
- use `sed` to escape double quotes in the above  env var. 

## Learning

- bash variable substitution and removal of new lines: https://stackoverflow.com/questions/19345872/how-to-remove-a-newline-from-a-string-in-bash
- `sed`
Had to take the source of the Orb and place it inline to find the issue with debug statements. This took a fair bit of experimentation to get to work.  
I've raised bugs on the orb for the first issue as i think it needs a bit of work.
- https://github.com/CircleCI-Public/slack-orb/issues/261
- 

## Checklist

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -106,37 +106,38 @@ pr_build: &pr_build
             pdf_docker_build,
             seeding_docker_build,
           ]
-    - ecr_scan_results:
-         name: ecr_scan_results_development
-         filters: { branches: { ignore: [master] } }
-         requires:
-           [
-             front_docker_build,
-             admin_docker_build,
-             api_docker_build,
-             pdf_docker_build,
-             seeding_docker_build,
-           ]
+    # - ecr_scan_results:
+    #     name: ecr_scan_results_development
+    #     filters: { branches: { ignore: [master] } }
+    #     requires:
+    #       [
+    #         front_docker_build,
+    #         admin_docker_build,
+    #         api_docker_build,
+    #         pdf_docker_build,
+    #         seeding_docker_build,
+    #       ]
 
-    - infrastructure_and_deployment/seed_environment_databases:
-         name: dev_seed_environment_databases
-         filters: { branches: { ignore: [master] } }
-         requires: [dev_environment_apply_terraform]
+    # - infrastructure_and_deployment/seed_environment_databases:
+    #     name: dev_seed_environment_databases
+    #     filters: { branches: { ignore: [master] } }
+    #     requires: [dev_environment_apply_terraform]
 
-    - infrastructure_and_deployment/run_cypress_test:
-         name: run_cypress_test
-         filters: { branches: { ignore: [master] } }
-         requires: [dev_seed_environment_databases]
+    # - infrastructure_and_deployment/run_cypress_test:
+    #     name: run_cypress_test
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -107,37 +107,37 @@ pr_build: &pr_build
...
+        name: run_functional_test
+        filters: { branches: { ignore: [master] } }
+        requires: [dev_seed_environment_databases]
 
     - slack_notify_domain:
         name: post_environment_domains
         filters: { branches: { ignore: [master] } }
         requires: [run_functional_test, run_cypress_test]
-
+        #requires: [dev_environment_apply_terraform, dev_account_apply_terraform]
```
townxelliot, line 140: Is this intentionally commented-out?

.circleci/config.yml
```diff
@@ -107,37 +107,37 @@ pr_build: &pr_build
...
```
Raffers, line 141: This line still needed?

scripts/pipeline/set_environment_variables/set_slack_post_environment_domains.sh
```diff
@@ -1,63 +1,68 @@
 #! /bin/bash
+    # needed as circleci did not santize the input for json properly.
+    SANITISED_COMMIT_MESSAGE=$(echo "${COMMIT_MESSAGE//$'\n'/\\\n}"  | sed 's/"/\\""/g')
```
townxelliot, line 3: Could you please give me a brief explanation of what's going on here as it's a bit opaque? I get the sed part but I'm not sure what's going on with the echo statement. Just for my own understanding, not a block on merging it.

Raffers (approved):

townxelliot (changes requested):
Just one question to address, in case comment is accidental.

.circleci/config.yml
williamfalconeruk: @Raffers - insurance policy just in case path to live has issues, i'll remove in my next PR i promise 😁

.circleci/config.yml
williamfalconeruk: yes - this was there during testing - see comment to @Raffers, i will remove in subsequent PR - wanted to switch back in case it failed in path to live.

scripts/pipeline/set_environment_variables/set_slack_post_environment_domains.sh
williamfalconeruk: The `echo` writes out to the shell variable with a substitution of actual new lines with a new line literal e.g. `\n'. Think of it as like sed command similar to `sed 's//$'\n'/\\\n/g`.

williamfalconeruk: To be fair it was a first attempt so could make this a bit clearer...

townxelliot (approved):

PR #436, opened 2021-04-27
LPAL-327 Move to opg-data-lpa Sirius gateway

townxelliot:
## Purpose

https://opgtransform.atlassian.net/browse/LPAL-327

Fixes LPAL-327

## Approach

Point service-api at the opg-data-lpa Sirius gateway rather than sirius-api-gateway.

This requires some careful review to ensure that the URLs used will be correct for live, and to make sure that the API responds as expected to our requests. This will impact on how statuses are reported for LPA applications in the dashboard.

## Learning

n/a

## Checklist


williamfalconeruk (dismissed):
@townxelliot - look like there might be some infra changes. I'll look into it and get back  to you.

williamfalconeruk:
just putting a hold on this temporarily while I find out if we have the right IAM permissions.

williamfalconeruk (approved):
Looks like this will be ok.

PR #449, opened 2021-05-05
LPAL-369 Remove browsersync and watch grunt plugins

townxelliot:
## Purpose

Remove browsersync to fix vulnerability in xmlhttprequest-ssl library used by grunt.

https://opgtransform.atlassian.net/browse/LPAL-369

Fixes LPAL-369

## Approach

browsersync is of questionable usefulness, is no longer maintained,
and relies on an XML HTTP library which has vulnerabilities
(see https://github.com/ministryofjustice/opg-lpa/security/dependabot/service-front/package-lock.json/xmlhttprequest-ssl/open).

Remove it.

Checked that `grunt build_js` and `grunt build_css` still work (they do).

The grunt tasks for browsersync and watch (the watch task is only used by browsersync) have been removed, along with the npm config for them.

Also cleaned up package.json slightly by ensuring lodash dependency is explicit.

## Learning

n/a

## Checklist


williamfalconeruk (approved):
Assuming that we're not using these options anywhere, i think this is good to go.

PR #450, opened 2021-05-06
LPAL-352 - revisit issue with slack notifications

williamfalconeruk:
## Purpose

revisit the slack issue where it fails to send. 

Fixes LPAL-352 (hopefully)

## Approach

remove variable expansion braces from `$COMMIT_MESSAGE` env var and associated sanitization.

## Learning

Var expansion not compatible with orb.

## Checklist

### DIFF
--- a/scripts/pipeline/set_environment_variables/set_slack_notify_production_release.sh
+++ b/scripts/pipeline/set_environment_variables/set_slack_notify_production_release.sh
@@ -54,7 +54,7 @@ generate_slack_notify_production_release()
                     [
                         {
                             "type": "mrkdwn",
-                            "text": "by user: ${CIRCLE_USERNAME} - branch: ${CIRCLE_BRANCH} - Commit Message: ${SANITISED_COMMIT_MESSAGE}"
+                            "text": "by user: ${CIRCLE_USERNAME} - branch: ${CIRCLE_BRANCH} - Commit Message: $SANITISED_COMMIT_MESSAGE"
                         }
                     ]
                 }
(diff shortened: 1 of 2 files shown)


scripts/pipeline/set_environment_variables/set_slack_notify_production_release.sh
```diff
@@ -1,10 +1,10 @@
 #! /bin/bash
     echo "COMMIT_MESSAGE:"
 
-    echo ${COMMIT_MESSAGE}
```
Raffers, line 4: So just looking into [brace-expansion](https://ss64.com/bash/syntax-expand.html) as I wasn't entirely sure on the difference. 
Are we sure it's not another character that's not escaped correctly causing this? 
As for looking at `{}` using this should have been fine, maybe in conjunction with the double quotes, but for the sanitised message it wouldn't have made a difference.

williamfalconeruk: Taking the braces out seems to fix the issue, it seems to treat some new lines as an overwrite of the message. it's something to do with the braces and the interaction with the shell.

williamfalconeruk: best to take the braces out as then it won't perform expansion in any case.

Raffers (approved):

scripts/pipeline/set_environment_variables/set_slack_notify_production_release.sh
Raffers: Ah, we've found another bug then 🙃

PR #459, opened 2021-05-12
LPAL-91 PDF unit test speed up

townxelliot:
## Purpose

Replace the several thousand lines of unit tests which are too slow to run (and hence turned off) with a few hundred which are faster (that we can run).

Coverage is good enough for now (85.2%), and the permutations we make assertions about are reasonably thorough. The main gap is in the Worker code, which isn't unit tested at all. However, this might be better done with integration tests, possibly followed by unit tests. I think the code here may need a serious refactor, so integration tests first makes more sense to me.

https://opgtransform.atlassian.net/browse/LPAL-91

Fixes LPAL-91

## Approach

The unit tests were previously removed from the code base as they were not being run. I reinstated several to get as close as I could to good coverage.

I then added service-pdf/tests/OpgTest/Lpa/Pdf/AbstractLp1Test.php to cover the gaps. To reduce the number of PDFs which had to be generated, I combined several tests into two groups, each of which only produces a single PDF. The first group (testPopulatePages) covers most of the gaps; the second group (testPopulatePagesAlternatives) covers cases which couldn't be covered in the first as they require contradictory data.

In addition, I reduced the assertions down to checking the data, strikethroughs and continuation pages which would be added to the PDF *by a particular test*. This removes the redundant testing over and over of the same data points which appear in all variants of the PDF, which is the approach the previous tests took.

While this increases the difficulty of working out which piece of code is tested in which group, it speeds up the test suite significantly, to the point where it runs in around 120 seconds on my machine.

I also added a PdftkFactory class to enable a specific pdftk implementation to be injected into the code. Previously, this was hard-coded to "pdftk". It took me a while to figure out why the tests failed on my machine, but I eventually tracked it down to the pdftk implementation on my Linux dev machine not being able to read the XPDF files produced by our code. By enabling the pdftk command to be passed to the tests, I was able to use the (working) Java implementation (as we use in live) instead.

## Learning

Plenty about PDF generation.

## Checklist


service-pdf/tests/OpgTest/Lpa/Pdf/Config/ConfigTest.php
```diff
@@ -0,0 +1,112 @@
...
+            'data2' => null,
+            'data3' => false,
+            5       => 'something',
+        ];
+
+        $data2 = [
+            'data1' => true,
+            'data2' => 'abc',
+            'data4' => 987,
+            5       => 'else',
```
Raffers, line 98: Should this value be `6` instead of `5`? Going by the assert listed below.

service-pdf/tests/OpgTest/Lpa/Pdf/ContinuationSheet2Test.php
```diff
@@ -0,0 +1,640 @@
...
+            'how'         => ReplacementAttorneyDecisions::LPA_DECISION_HOW_JOINTLY_AND_SEVERALLY,
+            'howDetails'  => 'test test',
+        ];
+
+        $content = $this->getHowWhenReplacementAttorneysCanActContent($lpa->document);
+
+        $pdf = new ContinuationSheet2($lpa, ContinuationSheet2::CS2_TYPE_REPLACEMENT_ATTORNEYS_STEP_IN, $content, 1);
+
+        //  Set up the expected data for verification
+        $formattedLpaRef = 'A510 7295 5715';
```
Raffers, line 232: I can see we're using this particular value a lot, should it be moved into `AbstractPdfTestClass` as a const?

service-pdf/tests/OpgTest/Lpa/Pdf/ContinuationSheet2Test.php
```diff
@@ -0,0 +1,640 @@
...
+        $templateFileName = 'LPC_Continuation_Sheet_2.pdf';
```
Raffers, line 233: Same again with this value. In fact, for all the good path tests values in the pdf moving these into `AbstractPdfTestClass` & then only over-ridding them when needed for bad path or testing the limits? Just so we can cut down a lot on the bloat in these files & make it a little easier to parse.

service-pdf/tests/OpgTest/Lpa/Pdf/Config/ConfigTest.php
townxelliot: According to the code, it shouldn't: the default behaviour for numeric keys (unless you pass `$preserveNumericKeys=true` to the `merge()` function, is to append the value to the list (which will assign it a new key). This seems unintuitive and not pleasant, but I'm not going to change it now as it may have knock-on effects elsewhere.

service-pdf/tests/OpgTest/Lpa/Pdf/ContinuationSheet2Test.php
townxelliot: I've gone through and removed as much repetition as I can.

service-pdf/tests/OpgTest/Lpa/Pdf/Config/ConfigTest.php
Raffers: I thought that might be the case, but you second guess yourself from looking at the code.

Raffers (dismissed):

service-pdf/tests/OpgTest/Lpa/Pdf/AbstractPdfTestClass.php
```diff
@@ -0,0 +1,249 @@
...
+
+use Opg\Lpa\Pdf\AbstractIndividualPdf;
+use Opg\Lpa\DataModel\Lpa\Lpa;
+use ConfigSetUp;
+use Opg\Lpa\Pdf\Aggregator\AbstractAggregator;
+use Opg\Lpa\Pdf\Config\Config;
+use Opg\Lpa\Pdf\PdftkFactory;
+use PHPUnit\Framework\TestCase;
+
+abstract class AbstractPdfTestClass extends TestCase
```
nickdavis2001, line 13: Should this class be named AbstractPdfTestCase rather than Class?  I am always wary of a class name that has the word class in it.... ;)

service-pdf/tests/OpgTest/Lpa/Pdf/AbstractPdfTestClass.php
```diff
@@ -0,0 +1,249 @@
...
+            foreach ($constituentPdfsForPage as $constituentPdfIdx => $constituentPdfForPage) {
+                //  Confirm that there is a corresponding entry in the expected values
+                if (!isset($expectedValue[$page]) || !array_key_exists($constituentPdfIdx, $expectedValue[$page])) {
+                    $this->fail(sprintf('Expected value missing while trying to verify constituent PDF index %s inserted after page %s', $constituentPdfIdx, $page));
+                }
+
+                $expectedPdf = $expectedValue[$page][$constituentPdfIdx];
+                $constituentPdf = $constituentPdfForPage['pdf'];
+
+                if ($constituentPdf instanceof AbstractIndividualPdf) {
```
nickdavis2001, line 181: I always worry when I see instanceOf .... feels like when that happens, the functions maybe ought to be on separate classes , and use inheritance/polymorphism to ensure the right function gets called for the class. However, this may be difficult to disentangle now I guess, assuming this is existing test code. Once for another story perhaps?

townxelliot: The problem (which I haven't fixed) is figuring out how to work out which PDFs are embedded in aggregator PDFs, which is bypassed here, effectively by ignoring them. In addition, some PDFs are included in the parent PDF as strings where the PDF doesn't need any data injected into it (e.g. a guidance page which is always the same). Rather than treat all constituent PDFs uniformly, we therefore have this mix of AbstractIndividualPdf objects, aggregator objects, and strings referencing the location of a PDF; which is why we therefore have this mess in the test.

Short of rewriting the whole thing to behave more sensibly and consistently, we have to keep this work-around to deal with that unsatisfactory data model.

nickdavis2001: Fair enough :)

nickdavis2001 (approved):

PR #461, opened 2021-05-17
LPAL-219 Cypress Admin Tests

Raffers:
## Purpose

Creates tests for the admin control panel & it's associated functionality 

Fixes LPAL-219

## Approach

Creates tests using the Cypress framework

## Learning

## Checklist


cypress/integration/Admin.feature
```diff
@@ -25,3 +25,44 @@ Feature: Admin
...
+
+  @focus
+  Scenario: View feedback sent to the service
+    Given I visit the admin sign-in page
+    And I log in to admin
+    And I click "feedback-link"
+    Then I am taken to the feedback page
+
+    # date fields for feedback range
+    Then I fill out "#id-day-start-date" element with "11"
```
nickdavis2001, line 54: Any chance of using a data-cy tag for this?  Or did you draw a blank trying to find where to put that?  I had a look myself and its not immediately obvious. ;)

Raffers: It was tried. The place to add it would be either
`service-admin/src/App/templates/app/feedback.phtml`
`service-admin/src/App/src/Form/Feedback.php`
`service-admin/src/App/src/Form/Fieldset/Date.php`

 But it's a field set that's pulled in, trying to applying attributes that aren't on the approved list doesn't work. I looked into trying to add in viewHelpers as we have on `service-front` but there is a fair amount of changes need to do this.

cypress/integration/common/i_fill_out.js
```diff
@@ -51,6 +51,14 @@ When("I force fill out", (dataTable) => {
             });
 });
 
+Then("I fill out {string} element with {string}", (element, value) => {
+   cy.get(element).clear({ force: true }).type(value, { force: true })
+});
+
```
nickdavis2001, line 57: Since this uses force, would you be willing to explicitly say "I force fill out" here? That is the pattern I did, to make it clear in the English text when we are using force and when not. Also, notice at the top of this file there is a step called "I type {string} into old style id {string}", which seems to me like a sibling to this but without the force. So maybe these steps could be named similarly?  To be honest I don't really like "old style id" and wished I called it something else. If you like I could rename that in a separate PR though.

cypress/integration/Admin.feature
nickdavis2001: Fair enough :)

nickdavis2001 (approved):
lgtm :)

PR #462, opened 2021-05-17
LPAL-92 Finer-grained return dates for LPAs

townxelliot:
## Purpose

https://opgtransform.atlassian.net/browse/LPAL-92

Fixes LPAL-92

## Approach

This PR uses the dispatchDate recently made available in the Sirius gateway, using that date (instead of registration date) + 15 days to give a more accurate idea of when an LPA should reach the citizen.

It also expands the dates used in the "Returned" LPA status, so that withdrawn, invalid or rejected date are used when available.

The integration tests I added are switched off in CI (as we currently can't seed data into Sirius in CI) but should work locally, where a mock Sirius is available as part of the docker-compose config.

## Learning

n/a

## Checklist


service-api/module/Application/src/Controller/StatusController.php
```diff
@@ -209,34 +196,58 @@ public function get($ids)
...
-                            $results[$lpaId] = ['found' => true, 'status' => $lpaDetail['status'], 'receiptDate' => $receiptDate, 'registrationDate' => $registrationDate, 'rejectedDate' => $rejectDate];
-                        }
-                        else if (isset($lpaDetail['status']) && $lpaDetail['status'] == $currentProcessingStatus) {
-                            $results[$lpaId] = ['found' => true, 'status' => $currentProcessingStatus,'receiptDate' => $receiptDate, 'registrationDate' => $registrationDate, 'rejectedDate' => $rejectDate];
-                        }
-                        else if(is_null($lpaDetail['status']) && !is_null($currentProcessingStatus) && is_null($rejectDate)){
-                            $results[$lpaId] = ['found' => true, 'status' => $currentProcessingStatus, 'rejectedDate' => null];
+                        if (isset($data['status'])) {
+                            // Data we only need if status is set already
+                            $data['receiptDate'] = isset($lpaDetail['receiptDate']) ? $lpaDetail['receiptDate'] : null;
```
townxelliot, line 222: Add private method for this

service-api/module/Application/src/Controller/StatusController.php
```diff
@@ -105,37 +105,18 @@ public function getCurrentProcessingStatus($id)
...
-        $applicationReturnDate = array_key_exists(LPA::APPLICATION_REJECTED_DATE, $metaData) ?
-            $metaData[LPA::APPLICATION_REJECTED_DATE] : $metaData[LPA::APPLICATION_REGISTRATION_DATE];
-
-        return $applicationReturnDate;
+        return array_key_exists(LPA::SIRIUS_PROCESSING_STATUS, $metaData) ?
+            $metaData[LPA::SIRIUS_PROCESSING_STATUS] : null;
     }
 
-    private function updateMetadata($lpaId, $lpaStatus, $receiptDate = null, $registrationDate = null, $rejectedDate = null)
+    private function updateMetadata($lpaId, $data)
```
townxelliot, line 119: Add issue to remove caching of metadata.

service-api/module/Application/src/Model/Service/ProcessingStatus/Service.php
```diff
@@ -178,10 +185,37 @@ private function handleResponse(ResponseInterface $result)
         if (isset($responseBody['rejectedDate'])){
```
townxelliot, line 185: Make a private method to reduce duplication of code.

service-front/module/Application/src/Controller/Authenticated/Lpa/StatusController.php
```diff
@@ -34,30 +36,64 @@ public function indexAction()
...
         $doneStatuses = array_slice($statuses, 0, array_search($lpaStatus, $statuses));
 
-        // Return either the applicationRejectedDate or the applicationRegistrationDate based on what's received from Sirius
+        // The metadata used here is stored in the db but is originally
+        // populated from Sirius.
         $metadata = $lpa->getMetadata();
 
+        // Return the rejected, invalid, withdrawn or dispatch date
+        // (whichever is latest). NB dates are strings at this point.
+        $candidates = [];
```
townxelliot, line 53: Cleaner way of doing this to avoid repetition.

service-front/module/Application/view/application/authenticated/lpa/status/index.twig
```diff
@@ -69,35 +69,31 @@
...
 
-                {% if returnDate != null  %}
-                    <p>We sent the LPA and letters by second class post on  {{ returnDate|date("d F Y")  }}. </p>
-                {% else %}
-                    <p>We sent the LPA and letters by second class post. </p>
-                {% endif %}
+                <p>The donor and all attorneys on the LPA will get a letter telling them the outcome.</p>
 
-                <p>If they have not arrived after 10 working days, please call 0300 456 0300.</p>
+                <p>If the LPA or the letters have not arrived by <span data-cy="lpa-should-receive-by-date">{{ shouldReceiveByDate|date('d/m/y') }}</span>, please call 0300 456 0300.</p>
```
townxelliot, line 94: `shouldReceiveByDate` could be null

service-front/module/Application/view/application/general/stats/index.twig
```diff
@@ -1,6 +1,6 @@
 {% extends 'layout/layout.twig' %}
 
-{%- set pageTitle = 'LPA Statistics' -%}
+{%- set pageTitle = 'LPA Statistics' -p%}
```
townxelliot, line 3: Typo

service-api/module/Application/src/Model/Service/ProcessingStatus/Service.php
townxelliot: On looking at this again, there's not a good way I can think of (short of passing both arrays into a function) to only set a key on $return if it is also set in $responseBody, so I'm going to leave this as-is.

Raffers (dismissed):

service-front/module/Application/view/application/authenticated/lpa/status/index.twig
townxelliot: If `shouldReceiveByDate` is null, the template handles it without an error, showing today's date. Given that it's highly unlikely a user could reach this page for an LPA with no dates set on it at all (which would result in a null processed date and therefore show today's date), this is a risk I'm willing to accept.

Raffers: I wouldn't assume so, worth checking. But I would assume this is populated when `received` is populated. As I think the logic should be once we have a received date we then populate `shouldReceiveByDate` with our current wait time info, which I think is 8 weeks?

nickdavis2001 (approved):
lgtm :)

PR #472, opened 2021-05-28
LPAL-60 SessionFactory cleanup

townxelliot:
## Purpose

Fixes [LPAL-60](https://opgtransform.atlassian.net/browse/LPAL-60)

## Approach

Original issue requested using dependency injection for the dynamodb client. On looking at the git history, I realised that we removed the dynamodb client from the SessionFactory anyway.

Instead, I took the opportunity to clean up the class slightly, and add a test to fully cover it in unit tests.

One point worth a mention is that I removed some code to set the domain on the cookie from the request. As far as I'm aware, this is redundant as PHP will default to setting the server domain as the cookie domain.

## Learning

n/a

## Checklist


williamfalconeruk (approved):
👍🏻

PR #477, opened 2021-06-03
Lpal 352 slack python notification

williamfalconeruk:
## Purpose

Alternative approach for creating slack messages, using Python, json serialization, Jinja and slack_python sdk
The aim here is to:
1. Make a more robust solution for sending slack messages using templates.
2. Make it reusable and simpler to use.
3. Make it available as a tool in it's own right called from the command line. (this will come later)

Fixes LPAL-352

## Approach
- Use the [Slack Block Kit Builder](https://app.slack.com/block-kit-builder) to format and design the json message templates
- convert them into Jinja templates, with the replacements needed
- Create a python script which:
    - Collects environment variables to use in the templates
    - Merges additional variables on the command line
    - Reads a template 
    - Renders it using the variables, with json string sanitization
    - Sends it as a message to a slack channel
 - Create an inline orb which takes:
    - When the message should be sent e.g. `on_success, on_fail, always` to mirror CircleCI statuses
    - Channel to send
    - Template file
    - Optional template folder - defaulted to `templates` 
    - Optional key values the the template can render in this format `key1=value1 key2=value2` etc..
 - rework the CircleCI to use an inline orb configuration
 - remove old scripts
 
Whilst i've use an inline orb to call the script and anchored it, the script could potentially be called in it's own right.

to use the script:

``` bash
cd scripts/pipeline/slack_notifications
pip install -r requirements.txt
chmod +x ./notify_slack_from_template.py
python ./notify_slack_from_template.py --template_file=<<template_file>>  --slack_channel=<<slack_channel>> --slack_token=<<slack_app_token>> --template_folder=<<template_folder>> --replacement_vars <<replacement_vars>>
```
Replace the angle bracket values with actual values environment variables as needed. you can omit some of these options (see above)
see the script itself for more details.

**Note:** Use of the above script requires you to have a slack app access token, with the right scopes attached. 
This is the same as the [original slack orb setup](https://github.com/CircleCI-Public/slack-orb/wiki/Setup). 

## Learning

- Python
- Python json library (`json.dumps()`, `json.loads()`)
- Jinja: https://jinja.palletsprojects.com/en/3.0.x/
- slack_sdk: https://slack.dev/python-slack-sdk/web/index.html#messaging
- argparse: https://docs.python.org/3/library/argparse.html
- anchors and aliases in yaml and CircleCI: https://circleci.com/docs/2.0/writing-yaml/
- DocString: https://www.geeksforgeeks.org/python-docstrings/

## Checklist


Raffers (approved):
lgtm 👌

.circleci/config.yml
```diff
@@ -654,26 +736,30 @@ orbs:
...
-                                        "${CIRCLE_WORKFLOW_ID}" \
-                                        "${approval_step}" \
-                                        "${CIRCLECI_API_KEY}"
+                    echo terraform log scan found no destructive changes. auto-approval will take place.
+                    exit 0
                 fi
 
+          - my_slack/notify:
+              when: on_fail
+              slack_channel: C01BDFQHTRU
```
johnnolan, line 751: Should the slack channel id be a env?

williamfalconeruk: perhaps this is a good idea. i'll look into it now @johnnolan. would make the channel id a secret as a project variable.

charlesmmarshall (approved):

PR #492, opened 2021-06-11
LPAL-175 Admin page: existing system message can't be removed

Raffers:
## Purpose

When I try to remove the system message, I am told that a value is required.

Deleting all of the message, followed by clicking ‘set system message’ should, and has previously, work.

Fixes LPAL-175

## Approach

Reorders logic surrounding setting the system message, adds an additional state when attempting to apply an empty message to system message

## Learning

## Checklist


williamfalconeruk (approved):
👍🏻

PR #501, opened 2021-06-15
LPAL-243 Replace obsolete libs and upgrade

townxelliot:
## Purpose

Fixes [LPAL-243](https://opgtransform.atlassian.net/browse/LPAL-243)

## Approach

* Update all libraries for service-front, service-api and service-admin.
* Upgrade service-admin to phpunit 9.
* Filter out cache adapters we don't need and which were giving deprecation warnings (could possibly do more here, but I left a couple I wasn't sure about).
* Remove deprecated/obsolete libraries as far as possible; container-interop can't be removed, as laminas depends on it; laminas-console can't be removed as laminas-test depends on it.
* Rework command-line scripts to use laminas-cli and rewrite tests.
* Remove unused script for dynamodb management.

## Learning

n/a

## Checklist


Raffers (approved):
lgtm,  removing more bloat!

The only question I have is about the removed section in the `composer.json` files, but it isn't a show stopper.

townxelliot:
> lgtm, removing more bloat!
> 
> The only question I have is about the removed section in the `composer.json` files, but it isn't a show stopper.

service-admin/composer.json?

Raffers:
> > lgtm, removing more bloat!
> > The only question I have is about the removed section in the `composer.json` files, but it isn't a show stopper.
> 
> service-admin/composer.json?
> 
> The other changes is composer.json files are just slight moving around of stuff, but the dependencies remain the same. I think this is a consequence of me doing `composer require laminas-cli`.

Ah, that would make sense. I wasn't sure if you had added other packages.

Also just noticed, I was talking about the `replace` section & not removed

townxelliot:
> > > lgtm, removing more bloat!
> > > The only question I have is about the removed section in the `composer.json` files, but it isn't a show stopper.
> > 
> > 
> > service-admin/composer.json?
> > The other changes is composer.json files are just slight moving around of stuff, but the dependencies remain the same. I think this is a consequence of me doing `composer require laminas-cli`.
> 
> Ah, that would make sense. I wasn't sure if you had added other packages.
> 
> Also just noticed, I was talking about the `replace` section & not removed

The "replace" section in effect tells composer to remove those dependencies. This is because we are using laminas-cache, but it pulls in every adapter, including some deprecated ones. By specifying the "replace", those dependencies are ignored.

PR #502, opened 2021-06-15
LPAL-412 Fix for potential session domain issue

townxelliot:
## Purpose

The LPAL-60 fix potentially broke session handling, so that
users who had recently logged into the front-end get a
"session expired" message. This could be because we changed the
domain on the lpa cookie, meaning that some users had multiple
cookies with the same name but overlapping domains. In this
situation, the old invalid cookie may have been checked by the
application, causing the session to appear as invalid.

This does suggest a possible bug in the application, as we
should ensure that we are checking the cookie for the correct
domain when deciding whether a session is invalid.

Fixes [LPAL-412](https://opgtransform.atlassian.net/browse/LPAL-412)

## Approach

Reinstate old code from the sessionmanagerfactory which explicitly sets the session domain.

Change name of session cookie to avoid collision with existing cookies (lpa -> lpa2).

## Learning

n/a

## Checklist

### DIFF
--- a/service-front/config/autoload/global.php
+++ b/service-front/config/autoload/global.php
@@ -58,7 +58,7 @@ return [
         'native_settings' => [
 
             // The cookie name used in the session
-            'name' => 'lpa',
+            'name' => 'lpa2',
 
             // Hash settings
             'hash_function' => 'sha512',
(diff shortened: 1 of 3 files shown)


townxelliot:
Putting this draft PR up in case we need to apply a fix if users start experiencing invalid session issues.

williamfalconeruk (dismissed):

williamfalconeruk (approved):

PR #505, opened 2021-06-21
LPAL-385 Improve LPA application status caching

townxelliot:
## Purpose

Fixes [LPAL-385](https://opgtransform.atlassian.net/browse/LPAL-385)

## Approach

* Fix broken integration tests (was checking for the wrong status on the detail page for one LPA; also selector was wrong, so we weren't looking for the status lozenge which was highlighted on the status detail page, just any lozenge with the right status, active or not)
* Check whether status and/or status dates (received, withdrawn, rejected etc.) changed - if either (or both) do, save metadata (previously, we only checked for status change)
* Better distinguish between the value returned to the client and the values used internally to decide whether to update metadata (e.g. use of "inDb" flag to mark whether a record for an LPA exists in the db)
* Significantly reduce number of SELECT queries used to render dashboard and check whether to save metadata: previously was N*2, where N is the number of LPA applications to show on the page; now 1.

Decision was to retain caching of status data in the db, even if this could mean the user sees stale data when Sirius is unavailable. Option to remove caching altogether was rejected, considered a worse user experience plus would need UX work to implement "loading"/"not found" messages to display status of data load from Sirius.

## Learning

Learned some more about Laminas db querying.

## Checklist


.idea/php.xml
```diff
@@ -219,13 +219,15 @@
...
+  <component name="PhpProjectSharedConfiguration" php_language_level="8.0">
+    <option name="suggestChangeDefaultLanguageLevel" value="false" />
+  </component>
   <component name="PhpUnit">
     <phpunit_settings>
       <phpunit_by_interpreter interpreter_id="4ed8db31-a6d2-4689-868b-5471721ce351" custom_loader_path="/app/vendor/autoload.php" phpunit_phar_path="" />
       <phpunit_by_interpreter interpreter_id="881ee65e-dab4-4d27-9858-db7669d088aa" custom_loader_path="/app/vendor/autoload.php" phpunit_phar_path="" />
       <phpunit_by_interpreter interpreter_id="64d769c9-e591-4551-b455-0f16f76b2498" custom_loader_path="/app/vendor/autoload.php" phpunit_phar_path="" />
-      <PhpUnitSettings custom_loader_path="$USER_HOME$/3_opg-lpa/service-pdf/vendor/autoload.php" phpunit_phar_path="" />
+      <PhpUnitSettings custom_loader_path="$USER_HOME$/3_opg-lpa/service-api/vendor/autoload.php" phpunit_phar_path="" />
```
nickdavis2001, line 230: Any idea why PHPStorm did this?  ;)

.idea/php.xml
```diff
@@ -219,13 +219,15 @@
       </server>
     </servers>
   </component>
-  <component name="PhpProjectSharedConfiguration" php_language_level="7.3" />
+  <component name="PhpProjectSharedConfiguration" php_language_level="8.0">
```
nickdavis2001, line 222: This looks OK, confirmed by Chris who is using PHPStorm more than me

service-front/module/Application/view/application/authenticated/lpa/status/index.twig
```diff
@@ -17,21 +17,21 @@
     {% if status != 'completed' %}
 
         <ol class="progress-bar">
-            <li class="progress-bar__steps {{ status == 'waiting' ? 'current-waiting' : '' }}">
+            <li class="progress-bar__steps {{ status == 'waiting' ? 'current-waiting" data-cy="lpa-progress-waiting' : '' }}">
                 <span class="{{ 'waiting' in doneStatuses ? 'progress-bar__steps--completed' : 'progress-bar__steps--numbers' }}"></span>
-                <span data-cy="lpa-progress-waiting" class="progress-bar__steps--text">Waiting</span>
+                <span data-cy="lpa-progress-text" class="progress-bar__steps--text">Waiting</span>
```
nickdavis2001, line 22: Good idea :)

.idea/php.xml
townxelliot: I'd happily take out the PHPStorm config. I just use it for debugging occasionally so I don't really care how it's configured. I don't think it should be in the repo at all to be honest.

.idea/php.xml
townxelliot: Maybe I'll just revert the changes to the .idea directory altogether.

.idea/php.xml
nickdavis2001: If it is not in the repo, how would we have shared config between the team though?  (Fair enough at the moment we may be trampling on each other )

nickdavis2001 (approved):
lgtm

PR #508, opened 2021-06-21
LPAL-404 Feedback cypress test failures

townxelliot:
## Purpose

Fixes [LPAL-404](https://opgtransform.atlassian.net/browse/LPAL-404)

## Approach

On investigation, error was tracked down to duplicate keys being inserted into the feedback table.

Fix was to remove explicit setting of ID when inserting records into this table, and ensure we reset the auto-indexing ID on this table when we clear it.

Also added some extra logging around db inserts for feedback.

## Learning

Could do with more logging at all levels around the application.

## Checklist


service-api/module/Application/src/Model/Service/Feedback/Service.php
```diff
@@ -26,15 +26,20 @@ class Service extends AbstractService
     public function add(array $feedback) : bool
     {
         // Filter out any 'non-allowed' fields
-        $allowed = ['agent', 'fromPage', 'rating', 'details', 'email', 'phone'];
+        $allowed = ['agent', 'fromPage', 'rating', 'details', 'email', 'phone', 'foo'];
```
williamfalconeruk, line 29: noticed `foo` here -is this right?

townxelliot: No it isn't! Well spotted, thanks.

Raffers (approved):

PR #511, opened 2021-06-23
LPAL-415 Downgrade mikehaertl/php-pdftk to 0.2.1

townxelliot:
## Purpose

Fix PDF being generated with squashed text for preferences and instructions.

Fixes [LPAL-415](https://opgtransform.atlassian.net/browse/LPAL-415)

## Approach

1. Generate PDF locally using current master
2. Downgrade php-pdftk library to version 0.2.1 which was current at the beginning of June, before merging work on LPAL-91 (where I upgraded the PDF generation dependencies)
3. Generate another PDF locally
4. Compare the two PDFs

When I downgraded the php-pdftk library, the output matches the desired output specified in bug LPAL-415. So downgrading this dependency in prod will hopefully have the same outcome.

## Learning

Really need to get away from php-pdftk. Also really need tests for visual appearance of PDFs.

One concern is caching of downloaded PDFs. I found that if I downloaded the PDF for a single LPA multiple times, I would get the cached copy which would not be regenerated. I had to create new LPAs to force a render. One concern is that a user might download a PDF again which has been cached recently and not see the fix.

## Checklist


williamfalconeruk (approved):

townxelliot:
Thanks for approving @williamfalconeruk. Are you OK with me pinning the versions of these libraries for now to prevent other unwanted changes? The php-pdftk library change shouldn't have altered how PDFs render (it was a minor version bump), but obviously has, so I'd like to prevent that happening in future. I'm similarly nervous about the other PDF library we're depending on.

williamfalconeruk:
> Thanks for approving @williamfalconeruk. Are you OK with me pinning the versions of these libraries for now to prevent other unwanted changes? The php-pdftk library change shouldn't have altered how PDFs render (it was a minor version bump), but obviously has, so I'd like to prevent that happening in future. I'm similarly nervous about the other PDF library we're depending on.

absolutely fine with this, though it would be good if we can see on which version it went wrong. alternatively we should get on with the change of library to a more modern one.

townxelliot:
> > Thanks for approving @williamfalconeruk. Are you OK with me pinning the versions of these libraries for now to prevent other unwanted changes? The php-pdftk library change shouldn't have altered how PDFs render (it was a minor version bump), but obviously has, so I'd like to prevent that happening in future. I'm similarly nervous about the other PDF library we're depending on.
> 
> absolutely fine with this, though it would be good if we can see on which version it went wrong. alternatively we should get on with the change of library to a more modern one.

Agreed, but it's a bit of a slow manual process to figure it out. I'd have to bisect between version 0.2.1 and 0.10 and make new LPAs each time I changed version (as the caching in the solution means you can't force regeneration of the PDF for an LPA, so I ended up making new ones to test the above fix).

PR #515, opened 2021-06-29
LPAL-291 Remove build artefacts from repo

Raffers:
## Purpose

The repo currently contains build artefacts which should instead be generated in both local and CI environments.

1. At the moment we lack a JS/CSS build in CI.  We should do this.
2. We should also provide a mechanism to build automatically locally when js files change. One option is to fire up a local docker container which watches for changes, and rebuilds when needed. Use team apparently have done this.
3. We can then remove all .min.js files from git. This will get use away from the problem where merge conflicts occur as result of changes-that-should-not-conflict, in the js files from which that is built.

Fixes LPAL-291

## Approach

Sets up the required step on CI to build the assets, along with local only build watch via grunt
Removes all safe to remove mini files from the repo

## Learning

https://circleci.com/docs/2.0/
https://gruntjs.com/using-the-cli
https://gruntjs.com/api/grunt
Delving slightly deeper into bash setting up the required scripts & containers

## Checklist


nickdavis2001 (approved):
lgtm

nickdavis2001 (dismissed):

williamfalconeruk (dismissed):

nickdavis2001 (approved):

PR #518, opened 2021-06-30
LPAL-340 auto clean up from psalm

nickdavis2001:
## Purpose

_Ran psalm against service-front, service-api, service-admin. _

## Approach

_PHP files auto-fixed by psalm_ 

## Learning

_system php is at 8, but psalm at the mo needs 7.4_

## Checklist


townxelliot:
Can I suggest we do this one component at a time (service-front first), so we can do it incrementally, as described in the original issue?

townxelliot:
Can I also cheekily suggest you remove the two commits which do some work then undo it? :)

PR #543, opened 2021-07-14
LPAL-451 reword and rejig post environment domain message

williamfalconeruk:
## Purpose

Our build pipelines take some time, but the infra is ready before the end to end Cypress tests are run. 

This change gives developers the URL's earlier so they can access the ephemeral environments. The post environment domains job will run as the tests start. 

We will also adjust our repo to wait on the jobs in Cypress to complete (that is an org infra change). See: https://github.com/ministryofjustice/opg-org-infra/pull/916

Fixes LPAL-451

## Approach

- PR build now completes on the tests with the message step done at the same time
- Clarifications to post environment domains content:
    - says environment deployment rather than ready
    - Caveat added about still running tests
    - link exposed to the running workflow for convenience 

## Learning

N/A

## Checklist


scripts/pipeline/slack_notifications/templates/post_environment_domains.j2
```diff
@@ -7,10 +7,24 @@
             "text":
             {
                 "type": "plain_text",
-                "text": "Online LPA Development Environment Ready",
+                "text": "Online LPA Development Environment Deployed",
                 "emoji": false
             }
-        }
+        },
+		{
```
townxelliot, line 14: Indentation is inconsistent here.

townxelliot (commented):
Only a minor comment, but looks fine other than that.

townxelliot (dismissed):

scripts/pipeline/slack_notifications/templates/post_environment_domains.j2
williamfalconeruk: i'll take a look :)

williamfalconeruk: yep  - it's placement of braces - another one of those potential things for code wars 🤣 . i'll fix and push - if you could re-approve please?

williamfalconeruk:
Example: 
Screenshot 2021-07-15 at 14 53 19

townxelliot (approved):

PR #545, opened 2021-07-16
LPAL-419 Performance platform worker lambda

townxelliot:
## Purpose

Fixes [LPAL-419](https://opgtransform.atlassian.net/browse/LPAL-419)

## Approach

Modify dev stack as follows:

* Add proxy lambda (function)
* Add worker lambda (docker container)
* Add SQS queue
* Use SQS queue as an event source for proxy lambda
* Forward events received at proxy lambda to worker lambda
* Worker lambda creates postgres connection and prints event JSON to stdout
* Client to make manual testing of the SQS - proxy - worker simpler

## Learning

* Initial approach, compiling psycopg2 statically and including in worker lambda function: https://github.com/jkehler/awslambda-psycopg2.
* Use a docker image for the worker lambda after realising static compilation is complicated: https://docs.aws.amazon.com/lambda/latest/dg/python-image.html
* Adding env vars to a lambda: https://docs.aws.amazon.com/lambda/latest/dg/configuration-envvars.html#configuration-envvars-config
* boto3 sqs client, used to trigger the SQS queue for testing: https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/sqs.htmlhttps://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/sqs.html
* Using SQS as an event source for a lambda: https://docs.aws.amazon.com/lambda/latest/dg/with-sqs.html#services-dynamodb-api
* localstack-client, Python library to make working with localstack from Python more straightforward: https://github.com/localstack/localstack-python-client
* Managing AWS lambdas from awscli (used in deployment scripts): https://docs.aws.amazon.com/cli/latest/reference/lambda/index.html

## Checklist


docker-compose.yml
```diff
@@ -1,5 +1,9 @@
 version: "3.7"
 
+networks:
```
townxelliot, line 3: Put comments in about why we have this network.

Raffers (approved):

PR #549, opened 2021-07-19
DLC enabled on pr branches

williamfalconeruk:
## Purpose

to enable docker layer caching on Cypress PR builds. 

Fixes LPAL-456

## Approach

- add a docker build step at start forcing build of layers, and enable caching
- switch docker layer caching on Cypress test job.

## Learning

## Checklist


williamfalconeruk:
closing due to benefit being less than the cost of adding this. we are also looking at GitHub Actions as an alternative and more modular approach.

PR #555, opened 2021-07-21
LPAL-459 Escape string used to query for LPAs

townxelliot:
## Purpose

Fixes [LPAL-459](https://opgtransform.atlassian.net/browse/LPAL-459)

## Approach

Use platform-specific quoteValue() method to escape the string
passed into the SQL statement used to query for LPAs by
donor name or LPA ID.

## Learning

n/a

## Checklist

### DIFF
--- a/service-api/module/Application/src/Model/DataAccess/Postgres/ApplicationData.php
+++ b/service-api/module/Application/src/Model/DataAccess/Postgres/ApplicationData.php
@@ -110,20 +110,19 @@ class ApplicationData extends AbstractBase implements ApplicationRepository\Appl
      */
     public function count(array $criteria) : int
     {
-        $sql    = new Sql($this->getZendDb());
+        $adapter = $this->getZendDb();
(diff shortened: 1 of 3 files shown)


williamfalconeruk:
just one question from me - is it worth adding a unit test for the specific issue we were seeing for this?

townxelliot:
> just one question from me - is it worth adding a unit test for the specific issue we were seeing for this?

I think that's possible, yes.

williamfalconeruk (approved):
Approved, if possible we could Unit test this. if not now, best to raise a ticket to cover off. mindful of the fix needing to go in.

townxelliot:
I'll look at providing a unit test, but happy to split that off into a separate ticket and put this into UAT now so it can go through.

townxelliot:
Slight issue with unit tests: there are no existing unit tests for the file in question! So I will put a unit test around at least that method as part of this PR.

PR #556, opened 2021-07-22
LPAL-462 ApplicationData unit tests

townxelliot:
## Purpose

Fixes [LPAL-462](https://opgtransform.atlassian.net/browse/LPAL-462)

## Approach

The ApplicationData class was fairly untestable when I first started, so I refactored it to make it more testable, creating a DbWrapper to better encapsulate db access. This could then be injected into the ApplicationData class, and mocked out in tests. The previous inheritance approach didn't give many benefits, as there were few shared methods in the parent class, so composition seemed like a better way to organise things.

I moved most of the code for SELECT queries into the DbWrapper class, including the code to do escaping of values in SELECT statements.

I added tests for DbWrapper->select().

I then added tests for the fetch() and count() methods in ApplicationData, which are simplified by moving a lot of the heavy lifting into DbWrapper.

**As this changes the fundamentals of db access, it's important to thoroughly test in UAT. In addition, this leaves the code in a half-and-half state, where ApplicationData has a different layout from the other \*Data classes. However, I think this structure is better, as it makes those classes potentially unit-testable.**

## Learning

Learned about the niceties of dealing with `yield` in PHP unit tests. 

Learned about using partial mocks which enable methods on a class under test to be patched. Very useful where it's difficult to mock out a dependency. For example, there's no easy way to construct a Sql object from an adapter; the Sql constructor requires you pass the adapter into it. This makes it difficult to mock. By encapsulating `new Sql($adapter)` in a `createSql` method, however, I could patch over `createSql` to return a mock Sql object. Much fun was had.

## Checklist

### DIFF
--- a/service-api/module/Application/src/Model/DataAccess/Postgres/AbstractBase.php
+++ b/service-api/module/Application/src/Model/DataAccess/Postgres/AbstractBase.php
@@ -3,6 +3,9 @@ namespace Application\Model\DataAccess\Postgres;
 
 use Application\Logging\LoggerTrait;
 use Laminas\Db\Adapter\Adapter as ZendDbAdapter;
+use Laminas\Db\Metadata\Source\Factory as DbMetadataFactory;
+use Laminas\Db\ResultSet;
+use Laminas\Db\Sql\Sql;
 
 class AbstractBase {
     use LoggerTrait;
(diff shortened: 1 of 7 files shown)


nickdavis2001 (approved):
lgtm

PR #565, opened 2021-07-28
Lpal 457 add infra and build details for perfplat worker

williamfalconeruk:
## Purpose

To add the infra and pipeline for the new performance-platform module in Make. We have based this on the infra in the Use an LPA team, but made it environment level as this means we do not share the lambda across ephemeral environments.

initially switched off by using a feature flag: `performance_platform_enabled` in `terraform/environment/terraform.tfvars.json`. 
We will commit a switch on when ready. 
**NOTE**: This is enabled for Development only at this time.

Fixes LPAL-457

## Approach

- create lambda infra and module
- add sqs infra
- update dockerfiles to have a CI element
- add CircleCI jobs and commands, and update workflows
- terraform feature flag
- tfsec scan

## Learning

See https://github.com/ministryofjustice/opg-use-an-lpa/pull/782 for examples used in this.

## Checklist


.circleci/config.yml
```diff
@@ -205,6 +216,24 @@ path_to_live: &path_to_live
         unit_test: false
         filters: { branches: { only: [master] } }
 
+    - build_containers_and_push_to_ecr/docker_build_lambda_image:
+        name: perfplat-lambda-docker-build-preproduction
+        account_id: "987830934591"
```
townxelliot, line 221: Should this be a configuration variable?

townxelliot (approved):
Looks OK to me so far as my limited knowledge of Terraform and AWS goes.

townxelliot:
Is there any way we can manually test the dev SQS and lambda once it's up, just to ensure the event source is working correctly.

.circleci/config.yml
williamfalconeruk: ideally, our counterparts on the Use team also do this in their CI config however. might be less of an issue on the migration to Github Actions.

williamfalconeruk:
> Is there any way we can manually test the dev SQS and lambda once it's up, just to ensure the event source is working correctly?

I'll look into this for you now.

PR #575, opened 2021-08-04
LPAL-475 Developer experience improvements

townxelliot:
## Purpose

[LPAL-475](https://opgtransform.atlassian.net/browse/LPAL-475)

Consists of various small fixes to get tests running better locally, and to correct a few issues with cypress feature files etc.

These changes were identified while working on LPAL-444 and originally formed part of that (larger) branch.

## Approach

n/a

## Learning

n/a

## Checklist


cypress/integration/common/fixtures.js
```diff
@@ -6,6 +6,7 @@ const {
 // but only if we're not running under CI. In CI we leave it intact for 
 // the next scenario, to simulate the user journey
 After({ tags: "@CleanupFixtures" }, () => {
+    cy.log('cleaning up fixtures');
```
nickdavis2001, line 9: Do we still want this in the log?

cypress/start.sh
```diff
@@ -46,9 +46,9 @@ fi
 
 
 RETVAL=$?
-echo printing $RETVAL
+echo "printing $RETVAL"
 
-if [ $RETVAL -eq 0 ]; then
+if [[ $RETVAL -eq 0 ]]; then
```
nickdavis2001, line 51: did that not work before then? bash syntax is weird ;)

nickdavis2001 (commented):
Just checking, the few savehandler files that were deleted? Is that definitely correct that they are no longer needed and/or will they get re-instated later?

townxelliot:
> Just checking, the few savehandler files that were deleted? Is that definitely correct that they are no longer needed and/or will they get re-instated later?

I can't find any reference to them anywhere. We are also using redis as our save handler, so (to me) it looks like that code is unused. I will take a second look and confirm, though.

cypress/start.sh
townxelliot: I got an error about "unary operator" with the other syntax. I checked the above works in both bash and zsh.

cypress/integration/common/fixtures.js
townxelliot: It doesn't hurt. It also makes it easier to see when the fixtures aren't being cleaned; without it, you just get an After > document log entry which doesn't explain where it comes from. I found it useful to see an explicit "I've cleaned up your fixtures" message.

PR #579, opened 2021-08-05
LPAL-365 New Feedback page for Performance platform in flask

nickdavis2001:
## Purpose

_New Feedback page for performance platform, implemented in Flask_

## Approach

_New flask app, using ideas from gov-frontend-jinja demo site,  nginx is configured to send specific traffic to this site instead of existing php site_ 

## Learning

_Learned more about negative lookahead in regexps, how to exclude patterns from regexps, see nginx.conf for examples of this being used. Learned about using map in nginx.conf, to add headers only if they aren't already provided by the upstream_

## Checklist


terraform/environment/ecs_front_v2.tf
```diff
@@ -0,0 +1,211 @@
...
+          "protocol" : "tcp"
+        }
+      ],
+      "volumesFrom" : [],
+      "logConfiguration" : {
+        "logDriver" : "awslogs",
+        "options" : {
+          "awslogs-group" : aws_cloudwatch_log_group.application_logs.name,
+          "awslogs-region" : "eu-west-1",
+          "awslogs-stream-prefix" : "${local.environment}.front-app.online-lpa"
```
williamfalconeruk, line 186: change this

terraform/environment/ecs_front_v2.tf
```diff
@@ -0,0 +1,211 @@
+//----------------------------------
+// front v2 ECS Service level config
+
+resource "aws_ecs_service" "front-v2" {
+  name                  = "front-v2"
+  cluster               = aws_ecs_cluster.online-lpa.id
+  task_definition       = aws_ecs_task_definition.front.arn
```
williamfalconeruk, line 7: this needs to be changed to v2 reference

terraform/environment/ecs_front_v2.tf
nickdavis2001: +1 Done

terraform/environment/ecs_front_v2.tf
nickdavis2001: Thanks. Sorted :)

cypress/integration/Authentication.feature
```diff
@@ -1,11 +0,0 @@
-Feature: Authentication
```
nickdavis2001, line 1: Remove this entirely as this simple login test is now duplicated all over the place

service-front-v2/app/__init__.py
```diff
@@ -0,0 +1,37 @@
...
+)
+
+app.jinja_env.trim_blocks = True
+app.jinja_env.lstrip_blocks = True
+
+csp = {
+    "default-src": "'self'",
+    "script-src": [
+        "'self'",
+        "'sha256-+6WnXIl4mbFTCARd8N3COQmT3bJJmo32N8q8ZSQAIcU='",
```
williamfalconeruk, line 27: what are these sha values? they look like secrets, do we need to store them in AWS?

williamfalconeruk (commented):
one question, which might need addressing.

nickdavis2001:
> I'd suggest the fonts are pulled in as part of the build step, rather than being committed into the repo.
> 
> Same goes for the CSS and JS which are part of the govuk-frontend: these should be pulled into the build and not committed to our repo.
> 
> Also, is the build.sh script called as part of our build anywhere? Or is it a convenience for developers? I'd like to see this incorporated into our build somehow if not.
> 
> Looks good to me apart from that.

Hi @townxelliot this is now sorted, it calls build.sh which pulls the fonts, css and js in, and these are not stored in our repo now

docs/architecture/0008-flask-site-security.md
```diff
@@ -0,0 +1,87 @@
...
+We will scan the python code
+
+### OWASP 2021 top 10 on flask site
+
+A01:2021 Broken Access Control :
+The initial release consists of static pages that are public. The first form will also be public. Ultimately, we aim to share session information with the php site. At this point tests will be required to ensure relevant content cannot be accessed without logging in.
+Except for public resources, deny by default.  This does happen to some extent in that the flask site needs nginx to actively be configured to proxy_pass pages to it. However we need to be careful with any wildcard instructions to nginx.
+
+A02:2021 Cryptographic failure, (formerly known as Sensitive Data Exposure).
+We avoid having any secrets in the repo.
```
williamfalconeruk, line 28: I would say this more forcefully, as this is MoJ policy:  `we do not keep any secrets in the repository and any in use will be protected using secrets management services, and accessed by code from there`. or along the lines of this. 
you may want to refer to this in general as well in this ADR: https://docs.opg.service.justice.gov.uk/documentation/guides/security_process.htmlhttps://docs.opg.service.justice.gov.uk/documentation/guides/security_process.html

williamfalconeruk (changes requested):
just one change to be a bit more forceful on secrets - see details in comment

docs/architecture/0008-flask-site-security.md
williamfalconeruk: you might want to add: `we use security scans and precommit hooks to check for secrets, in order to reduce the chance of these being committed`

nickdavis2001: All helpful :) Have added this.

ramchale (approved):

williamfalconeruk (approved):

PR #580, opened 2021-08-05
LPAL-476 - add codeql github action to capture analysis of python

charlesmmarshall:
## Purpose

Introduce a non-blocking github action that scans python code within the repo via codeql which will then report to the security tab

Fixes LPAL-476

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -1,0 +1,66 @@
+# For most projects, this workflow file will not need changing; you simply need
+# to commit it to your repository.
+#
+# You may wish to alter this file to override the set of languages analyzed,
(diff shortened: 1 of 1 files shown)


townxelliot:
Looks good @charlesmmarshall . However, how am I to interpret this page? https://github.com/ministryofjustice/opg-lpa/pull/580/checks?check_run_id=3250657375

charlesmmarshall:
@townxelliot the pr tries to compare to a scan against the default branch (master) to show what has been introduced / removed as part of it - but this is the first run so doesnt have a baseline to compare too just yet

If it finds any, it will show a table of what was changed and what it has introduced

townxelliot (approved):

PR #581, opened 2021-08-05
Lpal 467 psalm on api

charlesmmarshall:
## Purpose

Create a workflow to use psalm to scan the code within the php services

Excluding the pdf service at the moment, as the psalm command generates an error where it seemingly can't work out a class.


.github/workflows/psalm-admin.yml
```diff
@@ -0,0 +1,32 @@
...
+          php-version: '7.4'
+          tools: vimeo/psalm
+      - name: Composer install
+        run: composer install --prefer-dist --optimize-autoloader --no-suggest --no-interaction --no-scripts
+      - name: Run psalm
+        run: psalm --output-format=github --taint-analysis --report=results.sarif
+      - name: Upload Security Analysis results to GitHub
+        uses: github/codeql-action/upload-sarif@v1
+        with:
+          sarif_file: ./service-admin/results.sarif
```
townxelliot, line 32: Is this not relative to working directory?

townxelliot (approved):
I made one comment, but assuming it's just that I've misunderstood how the paths are resolved, happy to approve.

.github/workflows/psalm-admin.yml
charlesmmarshall: yeah, i rpesumed that oo, but it seemingly failed with just `./results.sarif`

happy to pop though a test though to see!

PR #595, opened 2021-08-18
LPAL-480 Cookie banner rework to GDS guidelines

townxelliot:
## Purpose

Fixes [LPAL-480](https://opgtransform.atlassian.net/browse/LPAL-480)

## Approach

Modify the cookie banner to use correct text and provide both accept and reject buttons. Add a save confirmation panel with "hide" button.

## Learning

n/a

## Checklist


williamfalconeruk (dismissed):
Looks good - understand why the caching was removed, after reading the commit for it. I  was going to ask but reasons make sense. the other bits look ok also.

williamfalconeruk (approved):

PR #602, opened 2021-08-20
LPAL-463/LPAL-523 Unit tests and refactor for Postgres data access classes

townxelliot:
## Purpose

Improve the testability of data access classes in service-api which execute queries against the database. Start filling out unit tests for these classes.

Fixes [LPAL-463](https://opgtransform.atlassian.net/browse/LPAL-463)

## Approach

Previously, I introduced a DbWrapper class to encapsulate (crudely for now) access to Postgres itself. This made it possible to start testing data access classes without a real database.

This PR injects a DbWrapper into all of the data access classes so they become testable (as the DbWrapper can be mocked/stubbed).

I suggest we put this into UAT (thorough!) to ensure that nothing has broken.

## Learning

These classes are still difficult to work with, as they construct and execute complex SQL queries. The table gateway metaphor doesn't really hold up, although this appears to be the original pattern. The Users table, for example, packs all of the user's address data, authentication tokens, contact details etc. into a document which is saved into a field as JSON. Consequently, the UserData class manages authentication, email address changes, physical addresses, authentication tokens and so on. This means it's a huge class with a lot of business logic embedded in it.

In addition, access to the tables is very raw, and uses arrays in a lot of places. This means that validation logic for queries is also shoe-horned into this class (as the objects representing table rows are raw result objects or even just arrays), resulting in methods with lots of conditional branching. This makes them harder still to test.

The presence of half a dozen mocks in some of these tests demonstrates the problem. I have struggled to come up with a better way of testing them, and have ended up writing some convenience methods and shortcuts for common testing patterns (e.g. result sets with X rows in).

While testing this, I also realised there are other parts of the code we don't exercise with our tests. For example, we don't have any tests which run the stats generation commands on the command line. I have manually tested generate-stats and account-cleanup tasks to ensure that the commits in this branch don't cause any issues for those commands.

## Checklist


service-api/module/Application/src/Model/DataAccess/Postgres/UserData.php
```diff
@@ -608,15 +595,12 @@ public function setInactivityFlag(string $userId, string $flag) : bool
      */
     public function getAccountsUnactivatedOlderThan(DateTime $olderThan) : iterable
     {
-        $sql    = new Sql($this->getZendDb());
-        $select = $sql->select(self::USERS_TABLE);
-
-        $select->where([
+        $where = [
             'active' => false,
             new Operator('created', Operator::OPERATOR_LESS_THAN, $olderThan->format('c')),
```
townxelliot, line 600: datetime format  - use DbWrapper::TIME_FORMAT

nickdavis2001 (approved):

service-api/module/Application/src/Model/DataAccess/Postgres/UserData.php
townxelliot: I've decided not to do this, to minimise changes which may not be required.

townxelliot:
@nickdavis2001 please re-review as I made a couple of extra changes for breakages in cron scripts, which I'd neglected to test thoroughly.

nickdavis2001 (approved):

williamfalconeruk (approved):

PR #615, opened 2021-09-01
LPAL-485/LPAL-444 CSRF fixes and load test foundations

townxelliot:
## Purpose

Fixes [LPAL-444](https://opgtransform.atlassian.net/browse/LPAL-444) and [LPAL-485](https://opgtransform.atlassian.net/browse/LPAL-485)

## Approach

Replace the default Redis save handler with a custom one which can mark some requests as being read only with respect to the session. This prevents slow handlers from incorrectly overwriting session data if they fail to execute before the client moves to a new page.

The load tests I've added create test users with an LPA for each, and check that there's no bleed between sessions, by ensuring that each user can only see their own single LPA when multiple users login in parallel. This should improve confidence that the custom save handler doesn't incorrectly share data between sessions and that its performance is sufficient to cope with multiple users logging in in parallel.

## Learning

Quite a bit about session handling in PHP.

## Checklist


townxelliot:
Would recommend we are careful with this: squash commits and watch logs when it goes live to ensure nothing screwy happens. Also needs downtime to merge, as it changes the session handling fundamentally.

townxelliot:
To be reviewed by team on 2021-09-06.

nickdavis2001 (approved):
lgtm

Raffers (approved):

PR #616, opened 2021-09-03
LPAL-486 - Cookie page not removing cookies

Raffers:
## Purpose

Saving preferences on the /cookies page doesn’t update the user’s current cookies correctly; e.g. if the user accepts usage cookies in the banner, then disables usage cookies on /cookies page, the Google Analytics cookies aren’t removed. This means that the user's preferences are not updated correctly.

Fixes LPAL-486

## Approach

Adds event handlers to radio buttons to set/remove cookies onClick

## Learning

## Checklist


service-front/assets/js/moj/moj.modules/moj.cookie-consent.js
```diff
@@ -11,6 +11,21 @@
                 window.GOVUK.cookie('cookie_policy') || window.GOVUK.setDefaultConsentCookie()
             }
 
+            if (this.isInCookiesPage()) {
+                var noJsMessage = document.querySelector('.no-js');
```
townxelliot, line 15: I think this probably won't have any effect. If JS is available, there won't be a no-js class, but a js-enabled one. So I think finding and removing this is redundant.

service-front/assets/js/moj/moj.modules/moj.cookie-consent.js
```diff
@@ -11,6 +11,21 @@
...
+                noJsMessage.classList.remove('no-js');
+
+                var cookieRadioYes = document.querySelector('input#usageCookies-yes');
```
townxelliot, line 18: I would suggest that the cookie consent is not saved until the user clicks the "Save cookie settings" button, in line with [guidance on the cookies page pattern in the design system](https://design-system.service.gov.uk/patterns/cookies-page/).

service-front/assets/sass/utilities/_helpers.scss
```diff
@@ -86,6 +86,9 @@
         display: none !important;
         visibility: hidden !important;
     }
+
+    display: unset !important;
```
townxelliot, line 90: Not sure we need this.

service-front/module/Application/src/Controller/General/CookiesController.php
```diff
@@ -23,32 +21,7 @@ public function indexAction()
...
-            $this->getResponse()->getHeaders()->addHeaderLine($newCookiePolicy->getFieldName(), $newCookiePolicy->getFieldValue());
-
-            $seenCookie = new SetCookie(self::SEEN_COOKIE_NAME);
-            $seenCookie->setValue('true')
-                ->setHttponly(false)
-                ->setSecure(true)
-                ->setPath('/')
-                ->setExpires(new \DateTime('+365 days'));
-            $this->getResponse()->getHeaders()->addHeaderLine($seenCookie->getFieldName(), $seenCookie->getFieldValue());
+            $this->flashMessenger()->addSuccessMessage('You’ve set your cookie preferences.');
```
townxelliot, line 24: Think this is OK, but it's perhaps misleading: clicking the button has no effect right now, as preferences are saved when clicking the radio buttons, so this flash message is just smoke and mirrors as a consequence. I'd prefer it if clicking the save button made a change to the page with JS, showing the success message, and requiring no trip to the server.

service-front/module/Application/view/application/general/cookies/index.twig
```diff
@@ -1,44 +1,50 @@
...
+
+    <div class="no-js" hidden>
         <p>
-            OPG and Government Digital Services use Google Analytics to measure how you use the service so we can improve it
-            based on user needs. Google Analytics sets cookies that store anonymised information about:
+            We cannot change your cookie settings at the moment because JavaScript is not running in your browser.
+            To fix this, try:
         </p>
-
         <ul class="list list-bullet">
```
townxelliot, line 41: These steps need to be carried out in order, so number them with `<ol>` rather than bullet them.

service-front/module/Application/view/application/general/cookies/index.twig
```diff
@@ -1,44 +1,50 @@
...
         {{ formElement(form.getCsrf) }}
 
-        <h2 class="heading-large">Cookie settings</h2>
+    <h2 class="heading-large">Change your cookie settings</h2>
 
-        <p>We use two types of cookie. One type is strictly necessary and the other type of cookie is optional.</p>
+    <p>We use two types of cookie. One type is strictly necessary and the other type of cookie is optional.</p>
 
-        <h3 class="heading-medium">Cookies that measure website use</h3>
+    <h3 class="heading-medium">Strictly necessary cookies</h3>
```
townxelliot, line 22: Suggest we raise an issue to modify the content so we enumerate the cookies in tables, as per the cookies page guidance.

townxelliot (changes requested):

service-front/module/Application/src/Controller/General/CookiesController.php
Raffers: Requires a rewrite of the flash box, this will need to be done under a separate ticket.

service-front/assets/js/moj/moj.modules/moj.cookie-consent.js
Raffers: Tested this locally, it is needed if we want the no js message appearing. The function in main only removes it from the body element & not from anything else.

service-front/assets/sass/utilities/_helpers.scss
Raffers: If we want the no js message appearing/disappearing correctly we need these values in the CSS

service-front/module/Application/view/application/general/cookies/index.twig
Raffers: Not really, depends on what's going on locally with their browser. These are only meant as a suggestion so they can look at what might be going on & not a definitive solution to what issues they are having locally.

townxelliot:
@Raffers given the complexity this is generating, I suggest we revert back to doing everything server-side. This would then simply mean amending the PHP code to add/remove the `_ga/_gid/_gat` cookies, effectively duplicating what the JavaScript code does (which it partially does anyway). Everything else can stay as is, meaning the user sees the alert on submit.

We can then look at reworking the alerts etc. and doing stuff wholly client-side at a later date.

That way we get a fix in for the bug without introducing too many changes at once.

townxelliot (changes requested):

service-front/module/Application/src/Controller/General/CookiesController.php
```diff
@@ -23,16 +22,20 @@ public function indexAction()
...
-                ->setHttponly(false)
-                ->setSecure(true)
-                ->setExpires(new \DateTime('+60 seconds'));
-            $this->getResponse()->getHeaders()->addHeaderLine($cookiePolicyViewed->getFieldName(), $cookiePolicyViewed->getFieldValue());
+            if ($form->get('usageCookies')->getValue() === 'yes') {
+                $cookiePolicy['usage'] = true;
+            } else {
+                $cookiePolicy['usage'] = false;
 
-            $cookiePolicy['usage'] = $form->get('usageCookies')->getValue() === 'yes' ? true : false;
```
townxelliot, line 35: Do we need to leave this line in place? I think this is how the form decides which radio button to set.

townxelliot (changes requested):

townxelliot (approved):

PR #627, opened 2021-09-09
LPAL-529 Fix Python unit tests and make them consistent

townxelliot:
## Purpose

Fixes [LPAL-529]()

## Approach

* Remove nose (test runner)
* Fix typo "retreive"
* Update use of pytest-cases (the code in main won't work with the version I installed, where the API is different)

## Learning

nose is deprecated

## Checklist


README.md
```diff
@@ -133,13 +133,23 @@ In this mode, the `Makefile` will fetch secrets using `aws secretsmanager` and `
...
+To run the unit tests for the PHP applications:
 
 ```bash
 make dc-unit-tests
 ```
 
-For how to run the functional tests, please see seperate README in tests/functional directory
+To run the unit tests for the performance platform (Python):
+
+```zsh
```
williamfalconeruk, line 144: could this be another make target? ok if not...

townxelliot: Could be, but would need to put in a docker container to do that. It could potentially be piggybacked on perfplatconfig, which only runs locally and configures perfplat for the local environment. For example, we could run the docker container, mount the volume and run the unit tests from within it. What do you reckon?

Raffers (approved):

PR #633, opened 2021-09-13
LPAL-497 Add policy to pdf and access logs

williamfalconeruk:
## Purpose

Security Hub finding fix to S3.
Enforces the use of SSL when accessing s3 buckets. we need to test this works with the PDF cache first before releasing.

Fixes LPAL-497

## Approach

Adds policy changes to`lpa_pdf_cache` and `access_logs`

## Learning

see: https://docs.aws.amazon.com/securityhub/latest/userguide/securityhub-standards-fsbp-controls.html#s3-5-remediation

## Checklist


nickdavis2001 (approved):
lgtm

PR #683, opened 2021-10-19
remove casper tests that are now duplicated by cypress

nickdavis2001:
## Purpose

_LPAL-273 / LPAL-366 remove casper tests that are now duplicated by cypress

## Approach

_cypress pdf tests now run on the fixtures that are already seeded in the DB, instead of as part of the stitched run. This gets around the pdf performance problem if we try to run it against the "kitchen sink" lpas that the stitched run creates.  Remove the old casper tests, the only casper tests that remain are the logout test, with things leading up to that. That will get removed soon, once the logout page is one of ours instead of on gov.uk_ 

## Checklist


williamfalconeruk (approved):
Looks good - nice clearout there @nickdavis2001 🗑️  😁

Raffers (approved):

PR #694, opened 2021-10-28
LPAL-448 - Lint PDF app

Raffers:
## Purpose

Improve code quality of the PDF app contained within Make LPA

Fixes LPAL-448

## Approach

Ran psalm & phpStan against the existing code base to fix the majority of issues, manually fixed the rest via the produced report

## Learning

## Checklist


service-pdf/tests/OpgTest/Lpa/Pdf/ContinuationSheet2Test.php
```diff
@@ -36,7 +36,7 @@ public function testConstructorExceptionContentPageNonNumeric()
         $this->expectExceptionMessage('The requested content page must be a positive integer');
 
         $lpa = $this->getLpa();
-        $contentPage = '1';
+        $contentPage = -1;
```
nickdavis2001, line 39: Quick Q - what does this change do?

nickdavis2001 (approved):
lgtm 🏆

williamfalconeruk (approved):

service-pdf/tests/OpgTest/Lpa/Pdf/ContinuationSheet2Test.php
Raffers: For transparency -
Nick Davis  09:59
PR looks good to me :) I made one comment for clarification of the 1 to -1 that was all (see comment)

Chris Rafferty  10:00
Php will auto caste it to a string, the fn needed a int for it to work.

Nick Davis  10:00
OK. why is is now minus 1 though instead of +1 ?

Chris Rafferty  10:02
The fn itself checks for a positive int, so needed to set it as a neg for the test to work correctly. As before it was passing it as a string and failing, now that would fail as a type error if it fails to caste it as an int.

Nick Davis  10:03
Gotcha :slightly_smiling_face:

PR #705, opened 2021-11-05
LPAL-593  LPAL-594 LPAL-595 LPAL-596 add vpc, KMS, Secrets and subnets to region state 

williamfalconeruk:
## Purpose

To add to the region based state file the following elements for eu-west-1:

- LPAL-593 - VPC State
- LPAL-594 - KMS
- LPAL-595 - Flow Logs
- LPAL-596 - Secrets

A migration script to perform these moves has been written and tested on Development. Also checked in here for reference, to be removed at a later date. We'll be applying this to the pre and prod environments in readiness for the next script, on approval.

We have not removed the account level resources as yet as we still need to do further work to change environment level and to add a build.

We've taken off the development switch on the Eu-west-1 module as this would meant a further state move would be required to take out the index reference. 

This is covered as a build step isn't in place yet for region.

**At this point we will need to sync any changes that are in account to the region state as and when, though new changes are unlikely to be made at this stage.**

We've also refactored locations of some account resources for clarity e.g. IAM.

## Approach
Note `tf` refers to this command: `aws-vault exec identity -- ` which is generally used as an alias.
- Initialise state for region and module `eu-west-1` using `tf init`
- Extract a script for creating the state moves
- Remove all items that are account level only.
- Script additions for backup of state files.
- Copy over (not move) into module account level resources that will need to be present in a different region. 
   - Using `tf state move` steps
   - All moves done in local state files, so as to not break remote state.
- Also noticed that the internet gateway was not actually stored as resource rather a data source. 
   -  this also has been scripted as a `tf show` step which is harvested
   -  then a `tf import` applied.
- Push into region state the changes made locally using `tf state push region.tfstate`  
  - we **don't** push the account state as this has to be removed in a later step.
- verify that the changes look sane on region state
   - i.e. that it does not attempt to recreate any resources in the migrated list using `tf plan`
- verify account level state to ensure no changes, using `tf plan`

## Learning

- Get id's out of terraform state: https://stackoverflow.com/questions/57811596/get-specific-value-out-of-the-terraform-state-file
- Terraform statefile inspection: https://www.terraform.io/docs/cli/inspect/index.html 
- Terraform state manipulation: https://www.terraform.io/docs/cli/state/index.html
- Terraform imports: https://www.terraform.io/docs/cli/import/index.html

## Checklist


charlesmmarshall (approved):

nickdavis2001 (approved):
lgtm

PR #723, opened 2021-11-19
Lpal 600 fix up ecs task

williamfalconeruk:
## Purpose

hook up ECS to new region level named resources.

Fixes LPAL-600

## Approach

- Add environment level values for short name into variables e.g. `dev, pre, prod`
- Hard code `eu-west-1` as region name - this will be refactored later when we modularise environment level resources.
- Point to new elasticache.
- Point to new PDF Cache S3 bucket.
- Add security group rule to allow access to new Elasticache, whilst temporarily retaining existing one.
- Move of data sources to a different file make this more canonical.

## Learning

N/A

## Checklist

### DIFF
--- a/terraform/environment/.terraform.lock.hcl
+++ b/terraform/environment/.terraform.lock.hcl
@@ -1,0 +1,60 @@
+# This file is maintained automatically by "terraform init".
+# Manual edits may be lost in future updates.
+
+provider "registry.terraform.io/hashicorp/aws" {
+  version     = "3.65.0"
+  constraints = "~> 3.0"
+  hashes = [
+    "h1:GCDkcISN83t+JK2U+ie3vaECnyxK0Sr6GjO7IrBOVeo=",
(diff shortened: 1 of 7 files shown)


Raffers (approved):

terraform/environment/.terraform.lock.hcl
```diff
@@ -0,0 +1,60 @@
+# This file is maintained automatically by "terraform init".
+# Manual edits may be lost in future updates.
+
```
nickdavis2001, line 3: Should this lock file definitely be committed?  If so , fine. If not, could it be added to .gitignore?

nickdavis2001 (approved):

terraform/environment/.terraform.lock.hcl
williamfalconeruk: commit it for now

PR #746, opened 2021-12-01
Lpal 632 upgrade php8 for admin, api, pdf,  bring in Alphagov/Pay locally for front

nickdavis2001:
## Purpose

_PHP 8 upgrade. Upgrade admin, api and pdf. Bring code for Alphagov/Pay into our local repo rather than in abandoned Alphagov repo, as groundwork for upgrading service-front as well_

## Approach

_Many version upgrades and a lot of retesting. I have brought in the Alphagov/Pay code as is, unchanged. Ideally we would change the namespace of this or possibly just use the application namespace, however to avoid changing too many things at the same time I would recommend that on a seperate PR once the dust has settled. Also for now, sticking with PHP 8.0 rather than recently released 8.1, because a number of these packages are only tested against 8.0. Nothing to stop us doing a follow-up PR to try bringing them up to 8.1, again once the dust has settled on this and happily run in prod for a week or so_ 

## Learning

## Checklist


service-front/composer.json
```diff
@@ -53,7 +52,8 @@
     "autoload": {
         "psr-4": {
             "Application\\": "module/Application/src/",
-            "MakeLogger\\": "../shared/logging/module/MakeLogger/src/"
+            "MakeLogger\\": "../shared/logging/module/MakeLogger/src/",
+            "Alphagov\\Pay\\": "module/Alphagov/Pay/src/"
```
williamfalconeruk, line 56: should this be a shared component? i guess not?

williamfalconeruk (dismissed):
just one question, but all good otherwise.

service-front/composer.json
nickdavis2001: This used to be in its own git repo, and handed from OPG to Alphagov in case anyone else wanted to use it. In the event no-one did, and it got abandoned. It was pinned at old dependency versions, so blocked upgrade to PHP 8. So as a first step I've pulled it in , as is. In the long run it could be pulled inside the Make code base and not be a standalone component. But that would require tweaking namespaces which I'd ideally prefer to do in seperate PR rather than change too much stuff at once,  if you guys concur?

Raffers (dismissed):

williamfalconeruk (dismissed):

Raffers (approved):

Raffers (approved):

williamfalconeruk (approved):

PR #759, opened 2021-12-13
log4j protection - set environment variable

nickdavis2001:
## Purpose

_log4j protection - set environment variable_

## Approach

_set env var in dockerfile_ 

## Learning

## Checklist

### DIFF
--- a/service-pdf/docker/app/Dockerfile
+++ b/service-pdf/docker/app/Dockerfile
@@ -7,6 +7,8 @@ RUN composer install --prefer-dist --no-interaction --no-scripts --optimize-auto
 
 FROM php:8.0-cli-alpine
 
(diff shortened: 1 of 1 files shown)


Raffers (dismissed):
Move it to `terraform/environment/ecs_pdf.tf` with all other system env's if we're going to use it.

williamfalconeruk:
> Move it to `terraform/environment/ecs_pdf.tf` with all other system env's if we're going to use it.

Given the seriousness of the exploit, we can push this in - and fix in ECS later. sorry to override @Raffers...

nickdavis2001:
For the record, keeping this in dockerfile rather than in terraform, would have ensured that this happened locally as well as in CI.  However, looks like this fix is no longer needed now

PR #789, opened 2022-01-24
LPAL-657 fix php-shellcommand to be latest 1.6.4 instead of 1.0.2

nickdavis2001:
## Purpose

_Upgrade to more secure version of php-shellcommand_

Fixes LPAL-657

## Approach

_Force version in composer.json_ 

## Checklist

### DIFF
--- a/service-pdf/composer.json
+++ b/service-pdf/composer.json
@@ -14,6 +14,7 @@
         "php": "^8.0",
         "aws/aws-sdk-php": "^3.204.5",
         "mikehaertl/php-pdftk": "0.2.1",
+        "mikehaertl/php-shellcommand": "1.6.4 as 1.0.2",
         "zendframework/zendpdf": "^2.0.2",
         "laminas/laminas-crypt": "^3.4.0",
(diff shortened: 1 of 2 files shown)


service-pdf/composer.json
```diff
@@ -14,6 +14,7 @@
         "php": "^8.0",
         "aws/aws-sdk-php": "^3.204.5",
         "mikehaertl/php-pdftk": "0.2.1",
+        "mikehaertl/php-shellcommand": "1.6.4 as 1.0.2",
```
williamfalconeruk, line 17: did it need to be redirected? or can this be a direct replacement?

nickdavis2001: I believe you have to explicitly override it. This command tells it to use 1.6.4 instead of 1.0.2.  If you just required 1.6.4 here, my (possibliy limited) understanding is that it could still decide to use 1.0.2

williamfalconeruk (approved):
might want to wait til the ECR fix is in, but LGTM 👍🏻

Raffers (approved):

PR #802, opened 2022-01-31
LPAL-610 RDS to Aurora Cluster support

williamfalconeruk:
## Purpose

Originially Sandbox the process to upgrade and migrate an RDS instance to an Aurora Global Cluster. We have kept the option in in case we need to upgrade to Aurora.

work on LPAL-610.

## Approach

- add global cluster option for aurora
- upgrade existing module to cater for the cluster options

## Learning

Having attempted to upgrade the cluster we worked out that there was quite a bit of cost  +(50%) to doing so especially  for  the write replica needed in another location, which may not be necessary for this service. 
Restoration times are in the hours for RPO and RTO, so we are looking into going back to the original approach of creating a snapshot of the RDS instance that is cross region replicated. Wil hold further discussions on this

We will then do some DR testing to try to stand this up in dev and then in preprod with the restructured infra. to follow.

## Checklist


nickdavis2001 (approved):

nickdavis2001 (approved):

williamfalconeruk:
not needed at the moment

PR #805, opened 2022-02-02
Lpal 420 create feedback database table, with associated infra  

nickdavis2001:
## Purpose

_Add perfplat table, both locally and in CI, done in such a modular way as to make the feedback component shareable as a unit in future_

## Approach

_Create another container similar to seeding, which is run locally from docker-compose , or in CI. Produce python script with sqlalchemy to make the table, and raw sql calls to grant and check privileges. Produce aws task using terraform and call this in CI using refactored python script from the one that starts seeding. _ 

## Learning

_I've made some shared python code into a module, this lives in scripts/pipeline/lib. It uses the setup.py to be installable as a module. Further refactoring meant that actually only 1 script, create_task, calls this, as the 2 callers were merged into 1. However, still left as an example of how to share python code as module_

## Checklist


Raffers (approved):

service-feedback-api/scripts/db/create_feedback_user.py
```diff
@@ -0,0 +1,64 @@
...
+    create_feedback_user()
+    if feedback_user_exists():
+        print('feedback user created succesfully.')
+    else:
+        print('ERROR : creation of feedback user appears to have failed.')
+
+if has_privileges():
+    print('feedback user already has privileges on perfplat table.')
+else:
+    grant_privileges()
```
williamfalconeruk, line 60: As discussed - we'll add further security checks to ensure we're not granting access to the user on other db resources. happy though for now as we're not yet using this table.

Raffers (dismissed):

terraform/environment/modules/environment/ecs_feedbackdb.tf
```diff
@@ -0,0 +1,97 @@
...
+  #tfsec:ignore:AWS007 - anything out
+  cidr_blocks       = ["0.0.0.0/0"]
+  security_group_id = aws_security_group.feedbackdb_ecs_service.id
+}
+
+//--------------------------------------
+// feedbackdb ECS Service Task level config
+
+resource "aws_ecs_task_definition" "feedbackdb" {
+  count                    = var.environment_name == "production" ? 0 : 1
```
williamfalconeruk, line 28: ah spotted another issue - this doesn't run in production

terraform/environment/modules/environment/ecs_feedbackdb.tf
```diff
@@ -0,0 +1,97 @@
...
+  execution_role_arn       = aws_iam_role.execution_role.arn
+  tags                     = merge(local.default_opg_tags, local.feedbackdb_component_tag)
+}
+
+
+//----------------
+// Permissions
+
+resource "aws_iam_role" "feedbackdb_task_role" {
+  count              = var.environment_name == "production" ? 0 : 1
```
williamfalconeruk, line 45: same - this doesn't run in production

williamfalconeruk (changes requested):
need to remove the non prod only condition as the table does run in prod

service-feedback-api/alembic/versions/e01fe3adad2e_create_perf_feedback_table.py
```diff
@@ -0,0 +1,30 @@
...
+down_revision = None
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.create_table(
+        'perf_feedback',
+        sa.Column('id', sa.Integer, primary_key=True),
+        sa.Column('rating', sa.String(2), nullable=False), # up to 2-letter acronynm , VS = very satisfied , etc
```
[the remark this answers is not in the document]
nickdavis2001: now changed to int

terraform/environment/modules/environment/ecs_feedbackdb.tf
nickdavis2001: fixed

williamfalconeruk (dismissed):
all good

Raffers (approved):

williamfalconeruk (approved):

PR #827, opened 2022-03-02
LPAL-674 Remove timestamp element of pay ref on fees page of LPA

Raffers:
## Purpose

To support work to be done in Sirius migrations, the the time stamp element of the pay reference number is to be removed from the fees page.

Only the A-ref is to be visible here. (The A-ref is to remain visible next to the barcode)

Fixes LPAL-674

## Approach

## Learning

## Checklist

<s>   I have updated documentation (Confluence/GitHub wiki/tech debt doc) where relevant</s>
<s>  I have added mandatory tags to terraformed resources, where possible</s>
<s>  If I added a package.json or composer.json, I also made sure this is included in the script in `.github/workflows/dependabot-update.yml`</s>

### DIFF
--- a/service-front/module/Application/src/Model/Service/Payment/Helper/LpaIdHelper.php
+++ b/service-front/module/Application/src/Model/Service/Payment/Helper/LpaIdHelper.php
@@ -1,9 +1,11 @@
 <?php
+
 namespace Application\Model\Service\Payment\Helper;
+
 class LpaIdHelper
 {
-    const LPA_ID_LENGTH = 11;
-    
+    public const LPA_ID_LENGTH = 11;
+
     /**
      * Helper function to construct the transaction ID
      * based on the LPA ID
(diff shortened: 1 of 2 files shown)


service-front/module/Application/src/Model/Service/Payment/Helper/LpaIdHelper.php
```diff
@@ -1,9 +1,11 @@
 <?php
+
 namespace Application\Model\Service\Payment\Helper;
+
 class LpaIdHelper
 {
-    const LPA_ID_LENGTH = 11;
-    
+    public const LPA_ID_LENGTH = 11;
```
nickdavis2001, line 7: Just wondered why that needed to change to public

Raffers: This was a precommit fix. But with PSR any `const` or `var` on a class should have their scope set.

williamfalconeruk (approved):

nickdavis2001 (approved):

PR #844, opened 2022-03-22
LPAL-701 email template changes, LPAL-702, LPAL-684 content changes on complete and checkout pages, LPAL-663 less logging

nickdavis2001:
## Purpose

_LPAL-701 email template changes, with LPAL-702, LPAL-684 content changes on complete and checkout pages_

## Approach

_Added Mr Green Knight seeded test data to aid testing. For LPAL-701, use the 3 new email templates for various conditions. For LPAL 702 and LPAL-684, Mods to twig templates. Many additions to unit tests to cover all 10 new scenarios. Removed excess now obsolete session loggin (LPAL-663)_ 

## Checklist


service-front/module/Application/src/Model/Service/Lpa/Communication.php
```diff
@@ -23,66 +24,59 @@ class Communication extends AbstractEmailService
...
+    private $to;
+    private $lpaTypeTitleCase;
+    private $userEmailAddress;
 
     public function sendRegistrationCompleteEmail(Lpa $lpa)
     {
         // Get the signed in user's email address.
-        $userEmailAddress = $this->userDetailsSession->user->email->address;
-        $to = [$userEmailAddress];
+        $this->userEmailAddress = $this->userDetailsSession->user->email->address;
```
townxelliot, line 36: I would avoid setting private member variables on a class unless I needed the state in multiple places. Even then, I would pass the state to the methods/functions which need it, rather than set member variables, unless I ended up with long argument lists.

Some of these are warranted, but I don't see any reuse of `$this->userEmailAddress` or `$this->to`, for example.

service-front/module/Application/src/Model/Service/Lpa/Communication.php
```diff
@@ -92,6 +86,114 @@ public function sendRegistrationCompleteEmail(Lpa $lpa)
...
+            'lpaTypeTitleCase' => $this->lpaTypeTitleCase,
+            'lpaPaymentReference' => $lpa->payment->reference,
+            'lpaPaymentDate' => $paymentDate,
+            'paymentAmount' => $amount,
+            'date' => $refundDate,
+        ]);
+
+        // If we have a separate payment address, send the email to that also
+        if (!empty($lpa->payment->email) && ((string)$lpa->payment->email != strtolower($this->userEmailAddress))) {
+            $this->to[] = (string) $lpa->payment->email;
```
townxelliot, line 117: Ah, there's a reuse of `$this->to`.

service-front/module/Application/src/Model/Service/AbstractEmailService.php
```diff
@@ -25,16 +25,17 @@ abstract class AbstractEmailService extends AbstractService
...
-    public const EMAIL_LPA_REGISTRATION_WITH_PAYMENT   = 'email-lpa-registration-with-payment';
-    public const EMAIL_NEW_EMAIL_ADDRESS_NOTIFY        = 'email-new-email-address-notify';
-    public const EMAIL_NEW_EMAIL_ADDRESS_VERIFY        = 'email-new-email-address-verify';
-    public const EMAIL_PASSWORD_CHANGED                = 'email-password-changed';
-    public const EMAIL_PASSWORD_RESET                  = 'email-password-reset';
-    public const EMAIL_PASSWORD_RESET_NO_ACCOUNT       = 'email-password-reset-no-account';
-    public const EMAIL_ACCOUNT_DUPLICATION_WARNING     = 'email-account-duplication-warning';
+    public const EMAIL_ACCOUNT_ACTIVATE                      = 'email-account-activate';
+    public const EMAIL_FEEDBACK                              = 'email-feedback';
+    public const EMAIL_LPA_REGISTRATION_WITH_PAYMENT1        = 'email-lpa-registration-with-payment1';
```
townxelliot, line 30: I'm not sure we need the numbered suffixes, as the names are distinct anyway. Alternatively, if you're trying to make clear they're a set, I'd call them `email-lpa-registration-1-with-payment`, `email-lpa-registration-2-with-cheque-payment` etc. to make the grouping clear.

service-front/module/Application/src/Model/Service/Lpa/Communication.php
```diff
@@ -92,6 +86,114 @@ public function sendRegistrationCompleteEmail(Lpa $lpa)
...
+                    'PTN' => true,
+                ]);
+        }
+    }
+
+    public function setUpEmailFieldsForPayments(Lpa $lpa)
+    {
+        // fill out email fields appropriately that apply to cheque and online payments
+        //
+        if (!empty($lpa->document->peopleToNotify)) {
```
townxelliot, line 155: I'd probably switch these clauses round so you have empty() first then !empty() (implied), rather than !empty() then (implied) empty().

service-front/module/Application/tests/Model/Service/Lpa/CommunicationTest.php
```diff
@@ -45,15 +46,90 @@ public function setUp(): void
         $this->service->setUserDetailsSession($userDetailSession);
     }
 
-    public function testSendRegistrationCompleteEmailWithoutPayment(): void
+    public function testSendRegistrationCompleteEmailWithoutPaymentButWithPersonToNotify(): void
```
townxelliot, line 49: Did you check coverage for the unit tests? There are a lot of conditionals? Might be worth making sure you caught all the branches.

townxelliot (changes requested):

service-front/module/Application/src/Model/Service/Lpa/Communication.php
nickdavis2001: Indeed :)

service-front/module/Application/src/Model/Service/Lpa/Communication.php
nickdavis2001: Both userEmailAddress and to are used in 2 places. I just made them member variables to avoid a load of passing arguments around.

nickdavis2001: I can pass userEmailAddress around though. will change this

service-front/module/Application/src/Model/Service/Lpa/Communication.php
nickdavis2001: +1. Done :)

service-front/module/Application/src/Model/Service/AbstractEmailService.php
nickdavis2001: @townxelliot the suffixes I simply matched what Amy called the templates within Notify.  I could change these if you really like, and have no objection, but then they either wouldn't match what Amy did or she'd have to rename the templates in Notify.  How much do you care about this?  ;)

service-front/module/Application/src/Model/Service/Lpa/Communication.php
nickdavis2001: OK , have changed userEmailAddress.  Have attempted to change $to, but it isn't very pretty because gets passed around. Might be better is $to stayed as a private variable

service-front/module/Application/tests/Model/Service/Lpa/CommunicationTest.php
nickdavis2001: I believe that all the cases where a payment is made, are covered by the unit tests.  For no payment being made, I've extended this to have seperate tests for Universal Credit, Receives Benefits, Awarded Damages,  in addition to the Low Income one. These extra tests are identical to the Low Income one except for the single param set on the payment .  I hadn't originally included them as expected behaviour was identical. However, adding them now protects us if we want  varying behaviour in future

service-front/module/Application/src/Model/Service/Lpa/Communication.php
nickdavis2001: Now fixed

service-front/module/Application/src/Model/Service/AbstractEmailService.php
townxelliot: Not enough to cause additional work.

townxelliot (approved):

townxelliot (approved):

townxelliot (approved):

PR #873, opened 2022-04-07
LPAL-723 Run as  non root user on Admin Containers 

williamfalconeruk:
## Purpose

Run Admin Containers as non root. 
This will act as a template for other containers with this as a basis for locking down root access.

Fixes LPAL-723 

## Approach

- Add non root user to container 
- `chown` copied files as this user
- Run containers as non root user 
- fix permissions on nginx to run as non root including folders

## Learning

- Copy files as non root: https://stackoverflow.com/questions/44766665/how-do-i-docker-copy-as-non-root
- Add user on alpine base images: https://stackoverflow.com/questions/49955097/how-do-i-add-a-user-when-im-using-alpine-as-a-base-image
- permissions issues with certain folders on nginx:
   - [nginx: [emerg] open() "/run/nginx.pid" failed (13: Permission denied)](https://stackoverflow.com/questions/64393237/nginx-emerg-open-run-nginx-pid-failed-13-permission-denied) 
   - [docker entrypoint running bash script gets "permission denied"](https://stackoverflow.com/questions/38882654/docker-entrypoint-running-bash-script-gets-permission-denied)
   - [How to run an nginx container as non root?](https://stackoverflow.com/questions/63108119/how-to-run-an-nginx-container-as-non-root)

## Checklist


service-admin/docker/app/Dockerfile
```diff
@@ -1,18 +1,23 @@
 FROM composer:2.1.11 AS composer
 
-COPY service-admin/composer.json /app/composer.json
-COPY service-admin/composer.lock /app/composer.lock
+RUN adduser -D -g '' user
+
+COPY --chown=user:user service-admin/composer.json /app/composer.json
+COPY --chown=user:user service-admin/composer.lock /app/composer.lock
 
 RUN composer install --prefer-dist --no-interaction --no-scripts --optimize-autoloader --ignore-platform-reqs
```
townxelliot, line 8: Is composer still running as root here? Do we need a su command or similar to switch to the new user? I see we have `USER user` further down, but presumably this part of the docker file won't be using that? I'd also suggest using a username other than "user" as it might get confusing. Maybe "serviceuser" or "servicerunner" or "svcuser" or something?

williamfalconeruk: @townxelliot all good questions: 
- Composer is still running root here, but as this is the compile part of a multistage build this will not be a problem. This is the "baking" phase so to speak for the vendor folders.
- The `user` is used for execution time only, so we can prevent running with root privilege.
- I think we can get around the root pip install by specifying a `--user` flag - but not tested that - maybe a separate ticket to look at that (in general).
- I'll fix the name to something more meaningful - it's a first attempt to get things working.

service-admin/docker/web/Dockerfile
```diff
@@ -1,12 +1,22 @@
 FROM nginx:stable-alpine
 
+RUN chown -R nginx:nginx /var/cache/nginx && \
+        chown -R nginx:nginx /var/log/nginx && \
+        chown -R nginx:nginx /etc/nginx/conf.d && \
+        touch /run/nginx.pid && \
+        chown -R nginx:nginx /run/nginx.pid
+
+RUN chmod -R 777 /etc/nginx/conf.d
```
Raffers, line 9: Should this be `777` instead of `774 or 770` I think it's a bit too broad otherwise. As we only want users & groups access & not others? If others do need access then it should be RO

williamfalconeruk: yep had same comment on the other one - will adjust as this change should read:
```suggestion
RUN chmod o+rwx /etc/nginx/conf.d
```

Raffers (approved):

nickdavis2001 (approved):

PR #902, opened 2022-04-21
LPAL-651 Prevent raw 500 page when newly-signed-up user changes email/password 

townxelliot:
## Purpose

Fixes [LPAL-651]()

## Approach

* Prevent RuntimeError breaking template rendering by instead returning empty data object
* Modify cypress scripts so we can sign up multiple users during test suite run
* Add test for mismatched password change and mismatched email change

## Learning

I didn't get to the bottom of why we were getting a runtime error when trying to get the LPA applications for the current logged-in user. However, I could see it happening. It's generally a bad idea to do this, though: better to return the equivalent of a [null object](https://en.wikipedia.org/wiki/Null_object_pattern) instead of breaking the template with an unexpected exception because an HTTP request failed. Also, shouldn't really be doing HTTP requests from view helpers, but that's another issue.

## Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -241,5 +241,5 @@ restitch:
 # this requires a globally-installed cypress
 .PHONY: cypress-open
 cypress-open:
-	aws-vault exec moj-lpa-dev -- python3 cypress/S3Monitor.py &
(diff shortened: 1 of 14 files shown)


Raffers (approved):

townxelliot:
This can be merged as it has passed UAT.

PR #938, opened 2022-05-20
LPAL-682 create terraform for cross region snapshot

williamfalconeruk:
## Purpose

create new submodule to be used to define the cross region backup for RDS. 

Fixes LPAL-682

## Approach

- Create environment
- Create module
- Add outputs from the Primary environment to allow this to be defined
note - we have existing backups in preproduction and production so these need importing into state. 

this will be in the form of
``` bash
terraform import "module.cross_region_backup[0].aws_db_instance_automated_backups_replication.instance" "<arn of db backup instance>"
```

## Learning

[terraform reference for aws db instance automated backups replication](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/db_instance_automated_backups_replication)

## Checklist


terraform/environment/terraform.tfvars.json
```diff
@@ -19,7 +19,7 @@
       "sirius_api_gateway_endpoint": "https://integration.dev.lpa.api.opg.service.justice.gov.uk/v1/lpa-online-tool/lpas/",
       "sirius_api_gateway_arn": "arn:aws:execute-api:eu-west-1:288342028542:*/*/GET/lpa-online-tool/*",
       "auth_token_ttl_secs": 4500,
-      "backup_retention_period": 0,
+      "backup_retention_period": 1,
```
williamfalconeruk, line 22:
```suggestion
      "backup_retention_period": 0,
```

terraform/environment/modules/environment/outputs.tf
```diff
@@ -29,3 +29,7 @@ output "feedbackdb_security_group_id" {
 output "cluster_name" {
   value = aws_ecs_cluster.online-lpa.name
 }
+
+output "aws_db_instance_arn" {
+  value = var.account.always_on ? aws_db_instance.api[0].arn : ""
```
nickdavis2001, line 34: Do you think that something like N/A or not set , would be better than just the ""  ?

williamfalconeruk: the blank has meaning (same as a null) - i'' try to set that now to null, good spot if it works :)
```suggestion
  value = var.account.always_on ? aws_db_instance.api[0].arn : null
```

nickdavis2001 (approved):

PR #940, opened 2022-05-23
LPAL-429 Remove Ordnance Survey key dependency

townxelliot:
## Purpose

[LPAL-429](https://opgtransform.atlassian.net/browse/LPAL-429)

## Approach

* Set the ORDNANCE_SURVEY env var to "-" in the Makefile.
* Modify the cypress tests so calls out to the OS postcode search are stubs (so we are still testing the UI elements involved in looking up an address from a postcode).

Still to be decided: How do we test the integration with the OS postcode lookup? Should we ping the service as part of our status page? Should we include verification of the response from OS to ensure their responses don't drift from our expectations? But these questions can be dealt with separately from LPAL-429.

## Learning

How to use cypress.intercept().

## Checklist


cypress/integration/UserPersonalDetails.feature
```diff
@@ -1,6 +1,15 @@
 Feature: Change User Personal Details
 
     As a user of Make an Lpa I want to be able to change my personal details
+
+    # NOTE: this test is not idempotent - you will get a failure the second time you run it
+    # TODO reset address in test data before running the test
+
```
nickdavis2001, line 7: Why does it fail 2cnd time and is there a way to avoid that happening?

README.md
```diff
@@ -50,31 +50,6 @@ For Linux users, revert to the instructions for installing phpcs, phpstan and pr
 * <https://phpstan.org/user-guide/getting-started>
 * <https://pre-commit.com/index.html#install>
 
-### Running locally without 3rd party integrations
```
mabely, line 53: How come we are removing this whole part of the readme? I thought they would still be relevant

cypress/integration/UserPersonalDetails.feature
townxelliot: Because the test changes the address and doesn't change it back again at the end. It would be better if we could poke the necessary data into the test before running it, but the lack of API access prevents that from happening.

README.md
townxelliot: Running without integrations does not work at present, because you have to have a Notify API key which is valid. The aim is to remove that dependency as we discussed this morning, so you could then run the test suite with no deps on external API keys etc.

cypress/integration/UserPersonalDetails.feature
nickdavis2001: Could the test be made to change it back at the end?  So as to leave things as they were?  Or is that too much of a pain?  (haven't looked at the code)

townxelliot: It would be possible, but not in scope for this ticket. I put the note in there to highlight to anyone checking this manually. However, could you please raise a ticket to improve this test as you explained?

nickdavis2001: OK fair enough :)

nickdavis2001 (dismissed):

mabely (dismissed):

mabely (approved):

PR #962, opened 2022-06-07
LPAL-803 Status detail page refactor and unit tests

townxelliot:
## Purpose

[LPAL-803](https://opgtransform.atlassian.net/browse/LPAL-803)

## Approach

* Refactor the service-front status page controller so that the view model can be created independently of it
* Add tests which create a view model (from stubbed out service-api responses + stubbed out database records), then render the "content" part of the status detail page, so we can assert about its content and test conditional sections etc.

## Learning

A little bit about how to use Twig.

## Checklist


townxelliot:
I will organise a PR party for this 2022-06-13 as it has some quite far-reaching changes.

PR #965, opened 2022-06-09
LPAL-754/LPAL-823 Upgrade service-admin and service-front to PHP 8.1

townxelliot:
## Purpose

[LPAL-754](https://opgtransform.atlassian.net/browse/LPAL-754)

[LPAL-823](https://opgtransform.atlassian.net/browse/LPAL-823)

## Approach

LPAL-754 (service-admin):

* Use psalm (level 4) to lint code for PHP 8.1.
* Ensure unit tests and cypress tests still run.

LPAL-823 (service-front):

* This only required a single code change (due to deprecation of DateTime accepting a null argument to its constructor).
* cypress test suite passes fine, as do unit tests.
* Will require thorough UAT to ensure there are no corners of the app which are affected by the upgrade but not covered by the cypress or unit tests.

## Learning

n/a

## Checklist


service-admin/src/App/src/Logging/LoggerTrait.php
```diff
@@ -0,0 +1,31 @@
+<?php
+
```
nickdavis2001, line 2: Wondering why we need a whole new file here?  Is it copied from elsewhere? I thought we had this in a seperate package that got pulled in by multiple projects?

townxelliot: We do, but that LoggerTrait is not usable in service-admin. This is because our LoggerTrait includes code specific to laminas-mvc, which we aren't using in service-admin. Using that LoggerTrait would therefore require adding a dependency on laminas-mvc to service-admin.

The shared LoggerTrait also expects certain headers to be on requests, and looks for events which aren't triggered in service-admin.

I did consider reworking the shared LoggerTrait so it could deal with both the laminas-mvc apps and service-admin. However, as the code is trivial, I decided to implement a simplified LoggerTrait in service-admin, rather than refactor.

nickdavis2001: 👍

nickdavis2001 (dismissed):

townxelliot:
Waiting for UAT

nickdavis2001 (approved):
Nice that you moved the logger out to shared area :)

townxelliot:
> Nice that you moved the logger out to shared area :)

It's still a different (simplified) logger, and it does mean that we have to have a laminas-mvc dependency in service-admin and service-pdf which they don't really need, but it is at least consistent and easier to follow.

PR #987, opened 2022-06-14
LPAL-803 Status detail page tests

townxelliot:
## Purpose

[LPAL-803](https://opgtransform.atlassian.net/browse/LPAL-803)

This is to provide unit testing for the status detail pages now that they are no longer tested by cypress (they were only tested locally anyway, so this is an improvement over our current CI pipeline for status pages anyway).

## Approach

* Refactor the service-front status page controller so that the view model can be created independently of it
* Add tests which create a view model (from stubbed out service-api responses + stubbed out database records), then render the "content" part of the status detail page, so we can assert about its content and test conditional sections etc.

## Learning

A little bit about how to use Twig.

## Checklist


nickdavis2001 (approved):
lgtm

PR #988, opened 2022-06-14
LPAL 724: Create SQS KMS keys

ntse:
## Purpose

Creates SQS queue-specific CMK KMS keys in preparation for changing the queues to use these keys.

Fixes LPAL-724

## Approach

Create KMS keys. This PR will be followed up by a change to the aws_sqs_queue resource to use these keys.

## Learning

n/a

## Checklist


terraform/region/kms.tf
```diff
@@ -1,3 +1,53 @@
+resource "aws_kms_key" "multi_region_pdf_sqs_encryption_key" {
+  enable_key_rotation = true
```
townxelliot, line 2: What's the key rotation period for this?

terraform/region/kms.tf
```diff
@@ -1,3 +1,53 @@
...
+}
+
+resource "aws_kms_alias" "multi_region_pdf_sqs_encryption_alias" {
+  name          = "alias/mrk_pdf_sqs_encryption_key-${terraform.workspace}"
+  target_key_id = aws_kms_key.multi_region_pdf_sqs_encryption_key.key_id
+  provider      = aws.eu-west-1
+}
+
+resource "aws_kms_replica_key" "multi_region_pdf_sqs_encryption_key_replica" {
+  description             = "PDF SQS encryption replica key"
```
townxelliot, line 14: Do replicas automatically rotate with the primary?

townxelliot (approved):

terraform/region/kms.tf
williamfalconeruk: the key rotation is defined in this document. https://docs.aws.amazon.com/kms/latest/developerguide/rotate-keys.html - it's 365 days, this should be seamless, and more than adequate for the needs of the pdf queue.

terraform/region/kms.tf
williamfalconeruk: https://docs.aws.amazon.com/kms/latest/developerguide/multi-region-keys-manage.html#multi-region-rotate yep they do :)

williamfalconeruk:
> Should we have a longer deletion window, as (I believe) 7 days is the minimum? Would this then cover situations where we know a key is going to be deleted and we might be scuppered if webops can't get to it within 7 days (e.g. if someone were on a 2 week holiday)?

i think we have enough cover from the WebOps CoP to reinstate these should these be accidentally deleted. I'd certainly wouldn't see this happen without someone knowing about it quickly. The standard is 7 days but we could bump this up if needed across all of our keys. one for another ticket perhaps?

PR #993, opened 2022-06-16
LPAL-840 Prevent 500 errors when user multiple-clicks the delete LPA button

townxelliot:
## Purpose

[LPAL-840](https://opgtransform.atlassian.net/browse/LPAL-840)

## Approach

* Show a flash message in the page if deletion fails (rather than throwing a runtime exception which displays as a 500 server error).
* Improve the JS handler for single use, so it disables event handlers the first time the element is clicked (rather than just disabling the element).
* Add a test for deleting an LPA (runs in parallel with other tests in CI).

## Learning

n/a

## Checklist


mabely (approved):

cypress/e2e/DashboardDeleteLpa.feature
```diff
@@ -0,0 +1,53 @@
+@DashboardDeleteLpa
+Feature: DashboardDeleteLpa
+
+    I want to be able to delete an LPA on the dashboard
+
+    Background:
+        Given I ignore application exceptions
+
+    @focus
```
mabely, line 9: What do we use `@focus` for?

mabely: I didn't find any indication of where it's used or much in docs. I asked Nick yesterday and it seemed that anything not tagged with it wouldn't run, but I didn't tag my feature test with it and it did run in CI

townxelliot: I think.

PR #997, opened 2022-06-17
LPAL-755 Upgrade service-pdf to PHP 8.1

townxelliot:
## Purpose

[LPAL-755](https://opgtransform.atlassian.net/browse/LPAL-755)

## Approach

See ticket.

## Learning

n/a

## Checklist


williamfalconeruk:
@townxelliot - this will need manual intervention and a merge before rebuilding. i'm closing temporarily until you are available.

service-pdf/src/Opg/Lpa/Pdf/AbstractLp1.php
```diff
@@ -732,6 +871,66 @@ private function getTrustAttorney(array $attorneys)
         return null;
     }
 
+    private function writeBarcodeToImageFile()
```
townxelliot, line 874: This is the main change which requires a review: changing the PDF rendering to use an image rather than a PDF. Once this is verified as working correctly, I will remove the code to render to a PDF. It's retained for reference until we can verify barcode image rendering works as intended.

service-pdf/src/Opg/Lpa/Pdf/AbstractLp1.php
```diff
@@ -22,6 +22,7 @@
 use Exception;
 use mikehaertl\pdftk\Pdf as Pdftk;
 use Laminas\Barcode\Barcode;
+use TCPDF;
```
townxelliot, line 25: TCPDF should be an explicit dependency

service-pdf/src/Opg/Lpa/Pdf/AbstractLp1.php
```diff
@@ -259,29 +341,50 @@ abstract protected function populatePageSix(PrimaryAttorneyDecisions $primaryAtt
...
             } else {
-                //  Add a strike through on the appropriate page
+                // Add a strike through on the appropriate page
                 $this->addStrikeThrough('people-to-notify-' . $i, 7);
             }
         }
 
         if (count($peopleToNotify) > self::MAX_PEOPLE_TO_NOTIFY_SECTION_6) {
-            //  TODO - Historic bug - the check box on the H&W PDF is named incorrecly
+            // TODO - Historic bug - the check box on the H&W PDF is named incorrecly
```
townxelliot, line 387: How old is this?

service-pdf/src/Opg/Lpa/Pdf/AbstractLp1.php
```diff
@@ -732,6 +872,58 @@ private function getTrustAttorney(array $attorneys)
...
+
+        // set image scale factor
+        $pdf->setImageScale(1);
+
+        $pdf->AddPage();
+
+        // filename, x position, y position, width, height, type, link, align, resize, dpi
+        $pdf->Image($barcodePngFile, 40, 789);
+
+        $barcodePdfFile = $this->getIntermediatePdfFilePath('barcode.pdf');
```
townxelliot, line 919: Re-check getIntermediatePdfFilePath returns a path unique to this PDF.

service-pdf/src/Opg/Lpa/Pdf/AbstractPdf.php
```diff
@@ -68,13 +68,17 @@ abstract class AbstractPdf extends PdftkPdf implements JsonSerializable
...
+     * @param string|null $templateFileName
      * @param array $options
      * @param ?PdftkFactory $pdftkFactory
      * @throws Exception
      */
-    public function __construct(Lpa $lpa = null, $templateFileName = null, array $options = [], ?PdftkFactory $pdftkFactory = null)
-    {
+    public function __construct(
+        Lpa $lpa = null,
+        $templateFileName = null,
```
townxelliot, line 78: ?string

service-pdf/tools/generate-pdfs-locally.php
```diff
@@ -1,43 +1,66 @@
 <?php
```
townxelliot, line 1: Rename directory to test-tools.

service-pdf/src/Opg/Lpa/Pdf/AbstractLp1.php
townxelliot: This is used in the old code for the generated barcode PDF, so we aren't doing anything new or unusual here.

service-pdf/src/Opg/Lpa/Pdf/AbstractLp1.php
townxelliot: This refers to a minor issue with a checkbox on the HW PDF having the wrong name, which this works around. It would require a change to the PDF to fix, which is not worth the hassle. I expanded the comment to better explain what's going on.

nickdavis2001 (approved):

PR #1010, opened 2022-06-23
LPAL-724 Use CMK for LB access logs bucket

ntse:
## Purpose

Switch from AWS-managed keys to CMK keys for access log
Fixes LPAL-724

## Approach

Use the recently created CMK KMS keys for the loadbalancer access logs S3 bucket

## Learning

n/a

## Checklist


terraform/region/kms.tf
```diff
@@ -1,7 +1,79 @@
+data "aws_kms_key" "access_log_key" {
+  key_id = "alias/mrk_access_logs_lb_encryption_key-${terraform.workspace}"
+}
+
+data "aws_iam_policy_document" "s3_loadbalancer_kms" {
```
townxelliot, line 5: Could you please explain why you need to enable root and key administrator permissions? Is it not possible to make a policy which only allows access from the LB (i.e. the third statement)? Or does the policy have wider application beyond the LB?

terraform/region/data_sources.tf
```diff
@@ -0,0 +1 @@
+data "aws_caller_identity" "current" {}
```
townxelliot, line 1: What's the purpose of this please?

terraform/region/kms.tf
```diff
@@ -1,7 +1,79 @@
...
+
+  statement {
+    sid    = "Allow ELB to use Key for Encryption"
+    effect = "Allow"
+
+    actions = [
+      "kms:Encrypt",
+      "kms:Decrypt",
+      "kms:ReEncrypt*",
+      "kms:GenerateDataKey*",
```
townxelliot, line 56: Does the ELB need to be able to describe keys and generate data keys? In what context would this happen?

townxelliot (changes requested):
I've added a few requests for clarification. It would probably help the team if you organised a PR party to explain CMKs, why we're using them, how they're scoped etc.

terraform/region/data_sources.tf
williamfalconeruk: this provides information to terraform for it's current aws context, which can be referenced by the configuration. See: https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/caller_identity

terraform/region/kms.tf
williamfalconeruk: GenerateDataKey : https://docs.aws.amazon.com/kms/latest/APIReference/API_GenerateDataKey.html - is required for the way that keys are generated, as there is a master key (CMK) and a data key which encrypts over the top, outside of KMS. this is usually stored with the data.
DescribeKey - https://docs.aws.amazon.com/kms/latest/APIReference/API_DescribeKey.html - allows the Principal to get information about a key.

townxelliot:
Thanks for clarifications @williamfalconeruk.

townxelliot (approved):

williamfalconeruk (approved):

terraform/region/kms.tf
```diff
@@ -15,6 +87,7 @@ resource "aws_kms_replica_key" "multi_region_access_logs_lb_encryption_key_repli
   deletion_window_in_days = 7
```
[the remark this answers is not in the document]
williamfalconeruk: This is to do with ensuring there is a minimum time left prior to actual deletion, as when you ask aws to do so it "marks" as deleted but retains the key for this period of time i.e. 7 days as specified here. Allows a recovery to happen if this was done by mistake.

PR #1015, opened 2022-06-27
LPAL-842 Improve cypress PDF test

townxelliot:
## Purpose

[LPAL-842](https://opgtransform.atlassian.net/browse/LPAL-842)

## Approach

* Pull in a library which can read PDFs in pure JS (pdf-lib)
* Add a checkPdf() cypress task which checks a PDF and returns true if it can be loaded, false otherwise
* Incorporate the checkPdf() call into the existing PDF download test (NB this replaces the check for the file size): if the PDF fails to load, the test fails; if it loads successfully, pass test if PDF has at least one page in it

## Learning

n/a

## Checklist

* [x] If I have a new OPG component dependency, I have updated the `metadata.json` with the repo location.


nickdavis2001 (approved):
lgtm

PR #1038, opened 2022-07-14
LPAL-907/LPAL-908 Set "cannot sign" for donor to value in reused LPA

townxelliot:
## Purpose

[LPAL-908](https://opgtransform.atlassian.net/browse/LPAL-908)

[LPAL-907](https://opgtransform.atlassian.net/browse/LPAL-907)

## Approach

See tickets

## Learning

n/a

## Checklist


cypress/start.sh
```diff
@@ -15,7 +15,8 @@ if [[ "$CYPRESS_CI" == "true" ]] || [[ "$CYPRESS_headless" == "true" ]] ; then
     echo "Running Cypress headless"
     # It's CI (used in CircleCI) or headless local CLI runs
     # so run the signup test first, then stitched, followed by all others, except for Clone which doesn't need a signup as it uses seeded user
-    if ! [[ $string =~ "StitchedClone" ]] ; then
+    if [[ "$CYPRESS_TAGS" =~ "not @IgnoreSignup" ]] ; then
```
nickdavis2001, line 18: I'm not a big fan of these double negatives and wonder if this could be refactored a bit.  Could we say here, if its StitchedPF or StitchedHW, then run sign up.? That'd mean no other tests have Signup run before them.  There may be test(s) in the. "Remaining tests" job that require signup, I can't remember.  If so, those could be in a seperate "Remaining Tests that Require Signup". (or something like that) job. Thats a better name than. "not ignore signup" I feel ....

nickdavis2001 (approved):

PR #1040, opened 2022-07-15
LPAL-843 Improve PDF caching approach

townxelliot:
## Purpose

[LPAL-843](https://opgtransform.atlassian.net/browse/LPAL-843)

NB ignore the commit history, as I based this on a branch which I then squashed while merging to main. (Consequently, the rebase on main was painful.) I'll squash this again when I merge it and fix the commit comment to remove the parts which are already on main.

## Approach

See ticket.

Note that this is very difficult to test: one would have to make a release, download the PDF, change the PDF generator in some obvious way, make another release shortly after the first, and download the PDF. With the old caching approach, the PDF would be the same for 24 hours after it was first generated; with this new approach, the new PDF should be available as soon as you do the second release.

## Learning

Our PDFs are cached in two mismatched ways:

1. Implicit cache refresh in S3: 24 hours after creation objects are removed
2. Explicit cache refresh based on the day: objects in the cache become invalid at 3 a.m. each day

This means that a cached PDF may be disposed of 3 hours after it's been created, even if the application hasn't changed; conversely, a PDF won't be removed from the cache even if the PDF generator changes until it's 3 a.m. or 24 hours have elapsed since it was created.

## Checklist


Makefile
```diff
@@ -87,7 +91,6 @@ dc-build-clean:
 	rm -fr ./service-front/public/assets/v2/js/vendor; \
 	rm -fr ./service-front/vendor; \
 	rm -fr ./service-pdf/vendor; \
-	if [ "`docker network ls | grep malpadev`" = "" ] ; then docker network create malpadev ; fi; \
```
townxelliot, line 90: Spotted these lines I missed in an earlier PR. These were to do with perfplat and are no longer needed so I cleaned them up while I was looking at the Makefile.

nickdavis2001 (dismissed):
lgtm

nickdavis2001 (approved):

PR #1047, opened 2022-07-19
LPAL-911 Improve cypress scripts

townxelliot:
## Purpose

[LPAL-911](https://opgtransform.atlassian.net/browse/LPAL-911)

## Approach

See ticket.

## Learning

n/a

## Checklist


cypress/s3_monitor.py
```diff
@@ -0,0 +1,184 @@
...
+
+    def assume_role_and_get_client(self):
+        sts = boto3.client(
+            "sts",
+            region_name="eu-west-1",
+        )
+
+        if self.in_ci:
+            print("S3Monitor starting, Assuming CI role")
+            role_arn = "arn:aws:iam::050256574573:role/opg-lpa-ci"
```
williamfalconeruk, line 31: ensure this is not hard coded here, as i think you might want to run this on the preproduction also.

cypress/s3_monitor.py
```diff
@@ -0,0 +1,184 @@
...
+        :param options: dict with v (verbose) and c (is in CI) boolean properties
+        """
+        self.verbose = options["v"]
+        self.in_ci = options["c"]
+        self.s3Client = self.assume_role_and_get_client()
+
+    def assume_role_and_get_client(self):
+        sts = boto3.client(
+            "sts",
+            region_name="eu-west-1",
```
williamfalconeruk, line 26: make this a parameter, will help with DR

cypress/s3_monitor.py
townxelliot: Where would the correct place be that I could get the strings for these roles from? Somewhere in terraform? Note that this is a conditional which switches between roles (and not something I touched), but you're right that it should be fixed.

cypress/s3_monitor.py
townxelliot: What's the correct value to use from terraform for this? Presumably I can pass an env var over which contains this.

PR #1052, opened 2022-07-21
LPAL-911 Improve cypress scripts

townxelliot:
## Purpose

[LPAL-911](https://opgtransform.atlassian.net/browse/LPAL-911)

## Approach

See ticket.

## Learning

n/a

## Checklist


cypress/stitch.py
```diff
@@ -0,0 +1,102 @@
+from pathlib import Path
+
```
nickdavis2001, line 2: What does replacing this in python buy us, which could be argued for this sort of task is less well-suited , and less readable, than bash/awk?  I can see that the original stitch.sh bash script had a lot of repetition, many lines almost the same, but it was clear to read, and if the repetition is really a problem, a loop could be used there.  Unless I am missing some other advantage of refactoring it this way?

townxelliot: Partly so it can be integrated easily with the other Python scripts. I'd argue the awk syntax is arcane and more difficult to read than the Python version. While you could do a loop in bash, the syntax is again archaic and hard to read. So for the sake of maintainability long term I'd argue Python is a better option.

nickdavis2001: Seems to me the problem with the original was it simply needed to loop over the list of files to stitch.

PR #1058, opened 2022-07-26
LPAL-871 Use opg-maintenance

ntse:
## Purpose

Use the new opg-maintenance service to display the maintenance page.
 
Fixes LPAL-871

## Approach

Updates the ALB to use the new opg-maintenance service.

## Learning

n/a

## Checklist


townxelliot:
This looks good @ntse. For my info, could you explain what triggers this to come into effect? How does the load balancer know when to apply the redirect?

townxelliot (approved):

townxelliot:
Just noticed that the cypress tests are being redirected to the maintenance page. I assume that means the condition which enables the maintenance page is not behaving as expected.

ntse:
> This looks good @ntse. For my info, could you explain what triggers this to come into effect? How does the load balancer know when to apply the redirect?

There's a script in docs/maintenance_mode which can activate and deactivate maintenance mode with the below commands. I had left it on whilst testing but forgot to disable it before the automated testing started. I'll rerun the tests and they should pass now.

./manage_maintenance.sh \                                
  --environment preproduction \
  --disable_maintenance_mode
  
  ./manage_maintenance.sh \                                
  --environment preproduction \
  --maintenance_mode.

PR #1062, opened 2022-07-28
LPAL-919 Allow feedback ci flask secret access

ntse:
## Purpose

Allow the Feedback CI user to read the Feedback Secret
Fixes LPAL-919

## Approach

As the Secret currently is manually created, the code performs a data lookup for the `opg-flask-api-token`. A `aws_secretsmanager_secret_policy` is then created referencing the ARN of the Secret.

## Learning
n/a
## Checklist

### DIFF
--- a/terraform/region/secrets.tf
+++ b/terraform/region/secrets.tf
@@ -188,7 +188,7 @@ data "aws_iam_policy_document" "opg_feedback_secrets" {
 
     resources = [
       data.aws_secretsmanager_secret.opg_flask_api_token.arn,
-      ]
+    ]
 
     actions = [
       "secretsmanager:GetSecretValue"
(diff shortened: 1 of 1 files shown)


terraform/region/secrets.tf
```diff
@@ -3,7 +3,7 @@
 # common
 resource "aws_secretsmanager_secret" "opg_lpa_common_admin_accounts" {
   name                           = "${local.account_name}/opg_lpa_common_admin_accounts"
-  tags                           = merge(local.default_opg_tags, local.admin_component_tag)
+  tags                           = local.admin_component_tag
```
nickdavis2001, line 6: Just wondered what the reason was for reducing the number of tags here?

ntse: The tags are duplicated at the provider level so Terraform "adds" these tags every time it runs. This just reduces noise in the Terraform output. I can remove this commit and add to a separate PR if necessary as it's only tangentially related to the main purpose of this PR

nickdavis2001: Make sense :). I'd keep it in this PR, why not?  Just wondered the exact reason that's all :)

charlesmmarshall (approved):

PR #1079, opened 2022-08-17
LPAL-949 tfsec ignore rule names instead of legacy id

ntse:
## Purpose

tfsec is no longer ignoring rules by their AWSXXX ID. 
Fixes LPAL-949

## Approach

Change all ignore rules to be rule names instead of legacy tfsec IDs.

## Learning

n/a

## Checklist


townxelliot (approved):
Looks straightforward, just changing ignore block numbers to constant strings, lgtm

PR #1082, opened 2022-08-19
LPAL-957 jQuery upgrade prep

townxelliot:
## Purpose

[LPAL-957](https://opgtransform.atlassian.net/browse/LPAL-957)

## Approach

Although full removal of jQuery will take some time, this PR lays some groundwork for that, namely:

* Include eslint as a dev dependency (JS linter)
* eslint config (useful for checking existing JS)
* Removal of unused JS modules
* Changes to node-build-assets to make it work more efficiently and consistently
* Minor fix to polyfilled details to remove double arrow in UI
* Remove jQuery from the dashboard JS module as a demonstration of the kind of thing required/to get the ball rolling
* Add tests for client-side validation in person form (name duplicates name elsewhere in application; age < 18; age > 100)
* Manually fix layout/style in one of the cypress support files

## Learning

n/a

## Checklist


service-front/assets/js/moj/moj.modules/moj.fees.js
```diff
@@ -1,24 +0,0 @@
...
-
-  moj.Modules.Fees = {
-
-    init: function () {
-      this.labelSubtextDisplay();
-    },
-
-    labelSubtextDisplay: function(){
-      $('input[name=reductionOptions]').change(function(){
-        $('.revised-fee').addClass('hidden');
```
townxelliot, line 15: None of these selectors appear in the PHP code or Twig templates, so this function does nothing.

service-front/assets/js/moj/moj.modules/moj.sticky-nav.js
```diff
@@ -1,44 +0,0 @@
...
-
-  moj.Modules.StickyNav = {
-
-    init: function () {
-      _.bindAll(this, 'stickIt');
-      this.stickIt();
-    },
-
-    stickIt: function () {
-      if ($('nav.progress-indicator').length) {
```
townxelliot, line 14: There are no HTML elements with the class progress-indicator, so this function will do nothing.

service-front/assets/js/moj/moj.modules/moj.user-timeout.js
```diff
@@ -1,105 +0,0 @@
...
-    gracePeriod: 1000 * 60 * 5, // ms * s * m = 5 minutes
-    timeout: null,
-
-    init: function () {
-      _.bindAll(this, 'warning');
-
-      var startTime = this.time();
-      // check to see if user is logged in
-
-      if (false) {
```
townxelliot, line 18: Note this condition never triggers, which suggests this is a redundant relic which was replaced with the session-timeout-dialog.js class.

service-front/assets/js/moj/moj.modules/moj.user-timeout.js
```diff
@@ -1,105 +0,0 @@
...
-
-    warning: function () {
-      var html = '',
-        self = this;
-
-      // TODO: get this HTML out of the JS into a template
-      // TODO: get someone to look at this copy, I made it up. Clive.
-      html += '<div class="timeout-body">';
-      html += '<h1>Session timeout</h1>';
-      html += '<p>You have been inactive for 55 minutes and will be automatically logged out in 5 minutes.</p>';
```
townxelliot, line 45: We don't ever see this hard-coded text in the app, which suggests it is unused. We also have a cypress test for the session timeout dialog box which passes after this code is removed.

cypress/e2e/common/i_fill_out.js
```diff
@@ -1,17 +1,17 @@
 import { When, Then } from "@badeball/cypress-cucumber-preprocessor";
```
townxelliot, line 1: I manually linted this file as the spacing was all over the place.

service-front/assets/js/moj/moj.modules/moj.who-are-you.js
```diff
@@ -1,27 +0,0 @@
...
-    init: function () {
-      this.selectionBehaviour();
-    },
-
-    selectionBehaviour: function(){
-      $('#WhoAreYou input:radio:checked').parents('.set').find('.subquestion:first').show();
-
-      $('#WhoAreYou input:radio').change(function (){
-        // Hides all sub-questions...
-        $('#WhoAreYou .subquestion').hide();
```
townxelliot, line 18: There are no HTML elements with the subquestion class, so this will never fire. Hence more redundant code.

service-front/assets/js/opg/jquery-plugin-opg-hascrollbar.js
```diff
@@ -1,14 +0,0 @@
...
- * JQuery Postcode Lookup plugin for OPG-LPA project
- * Relies on /postcode/lookup route
- *
- * @copyright The Engine Group
- * @author Chris Moreton <chris.moreton@netsensia.com>
- */
-
-(function ($) {
-  'use strict';
-  $.fn.hasScrollBar = function () {
```
townxelliot, line 11: The hasScrollbar method is never called anywhere, so this is redundant.

service-front/assets/sass/patterns/_details.scss
```diff
@@ -9,3 +9,12 @@ details {
         margin-top: em(10, 20);
     }
 }
+
+// remove the arrow added by the govuk CSS :before selector
+// on summary elements inside a polyfilled <details>
+[data-polyfilled-details=true] .govuk-details__summary:before {
```
johnnolan, line 15: Very picky but this could also be like this to make it more modular

`
[data-polyfilled-details=true] .govuk-details__summary {
    padding-left: 0 !important;
    
   &:before {
    display: none !important;
   }
}`

johnnolan (approved):
LGTM - nothing I can see with my eyes

service-front/assets/sass/patterns/_details.scss
townxelliot: Yeah, but I hate SASS

townxelliot: I mean, I'm being flippant there, but I think Sass can make people lazy. Rather than thinking about connecting styles directly to markup, perhaps by marking elements with their role and associating CSS selectors with that, it can result in horribly nested selectors which are difficult to follow.

I would also argue it obfuscates findability. As the selector is split up across declarations (as here, where the :before is on an & symbol), you can't search directly for the `.govuk-details__summary:before` selector in your Sass files.

Finally, it pulls in a load of additional dependencies (Ruby) and needs to be compiled all the time during development, requiring a watcher process etc.

I have added the styles into the Sass file here, but only because that's how the project is set up. If I had worked on the project from the start, it would probably not have any Sass files in it at all!

johnnolan (approved):

PR #1090, opened 2022-08-30
LPAL-918 pull feedbackdb from opg-feedback repo instead of building it

nickdavis2001:
## Purpose

_Remove from opg-lpa, the building of feedbackdb, which now duplicates what is done in opg-feedback_

## Approach

_Now that feedbackdb is built by opg-feedback repo seperately, instead of continuing to build this in opg-lpa  we now pull a pre-built image from opg-feedback ECR repo and use that_

## Checklist


terraform/environment/modules/environment/ecs_feedbackdb.tf
```diff
@@ -56,7 +56,7 @@ locals {
     {
       "cpu" : 1,
       "essential" : true,
-      "image" : "${data.aws_ecr_repository.lpa_feedbackdb_app.repository_url}:${var.container_version}",
+      "image" : "${data.aws_ecr_repository.lpa_feedbackdb_app.repository_url}:latest",
```
townxelliot, line 59: Shouldn't this be tied to a specific version?

nickdavis2001: If we leave it at latest, then we keep up to date with what is built and automatically tested in the opg-feedback repo.  If we pin it to a version, then (1) we have to actively keep up to date with it, when e:g it receives any security patches and (2) opg-feedback then has a dependency that it cannot delete old versions of images as opg-lpa depends on them.  Do you think there are advantages in pinning it that outweigh that?

townxelliot: My only concern is accidentally releasing the wrong version of the feedbackdb, e.g. if we roll back to a previous deployment of another component which only works with a particular version of feedbackdb.

townxelliot (approved):

PR #1103, opened 2022-09-14
LPAL-853 Create PG13 parameter group

ntse:
## Purpose

Create Postgres 13 parameter group to be used by PG13 RDS instances.

Fixes LPAL-853

## Approach

Create a separate PG parameter group that will exist alongside the current parameter group.

## Learning

## Checklist


charlesmmarshall (approved):

terraform/environment/modules/environment/rds.tf
```diff
@@ -132,6 +132,30 @@ resource "aws_db_parameter_group" "postgres-db-params" {
...
+  family      = var.account.psql13_parameter_group_family
+  parameter {
+    name         = "log_min_duration_statement"
+    value        = "500"
+    apply_method = "immediate"
+  }
+
+  parameter {
+    name         = "log_statement"
+    value        = "all"
```
townxelliot, line 147: Do we need to log all statements? I imagine that's going to produce massive logs.

terraform/environment/modules/environment/rds.tf
```diff
@@ -132,6 +132,30 @@ resource "aws_db_parameter_group" "postgres-db-params" {
...
+    apply_method = "pending-reboot"
+  }
+
+
+  parameter {
+    name         = "rds.log_retention_period"
+    value        = "1440"
```
townxelliot, line 154: Though if we're only keeping them for 24 hours, probably OK

townxelliot (approved):

townxelliot:
@ntse looks good, just wanted to check whether we are using the same logging config as other teams (retain logs for 24 hours, log all statements).

terraform/environment/modules/environment/rds.tf
ntse: Although it will all SQL statements, it will only log it if the transaction takes longer than 500 ms (specified by log_min_duration_statement)

PR #1107, opened 2022-09-15
LPAL-978 Fix Cleanup environment

ntse:
## Purpose

Add missing env variable for TF destroy in environment cleanup

Fixes LPAL-978

## Approach

Add TF_VAR_pagerduty_token to Terraform

## Learning

## Checklist

### DIFF
--- a/.github/workflows/workflow_terraform_environment_cleanup.yml
+++ b/.github/workflows/workflow_terraform_environment_cleanup.yml
@@ -53,5 +53,7 @@ jobs:
 
       - name: Destroy PR Terraform Workspaces
         working-directory: ./terraform/environment
+        env:
+          TF_VAR_pagerduty_token: ${{ secrets.PAGERDUTY_TOKEN }}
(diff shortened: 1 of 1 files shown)


.github/workflows/workflow_terraform_environment_cleanup.yml
```diff
@@ -53,5 +53,7 @@ jobs:
 
       - name: Destroy PR Terraform Workspaces
         working-directory: ./terraform/environment
+        env:
```
townxelliot, line 56: Was it just the case that the script failed because it couldn't communicate with pagerduty?

townxelliot (approved):

.github/workflows/workflow_terraform_environment_cleanup.yml
ntse: Yeah, the terraform won't run without a pagerduty_token Terraform variable being passed to it

PR #1125, opened 2022-09-29
LPAL-844 add eslint for linting javascript in opg-lpa repo

nickdavis2001:
## Purpose

_LPAL-844 add eslint, initially configured with a few rules turned off to ignore those errors for now, with ultimate plan to fix errors in time_

## Approach

- add eslintrc.json. this extends eslint:recommended, which is a little more thorough than standard. 
- a few rules in there to get it to pass, which we will take out as we fix the code in future.  
- babel config file. 
- mods to package.json to get eslint and configure it to lint specifically the application javascript and the cypress steps (turns out that type of errors reported are similar for both). 
- little bit of re-ordering in package.json
- we are NOT for now bothering with prettier that opg-pdf-service uses , as trying that gave huge error numbers. we may perhaps do that later once we've fixed the main errors_

## Learning

_Useful although very thorough and long example of eslint config here : https://github.com/standard/eslint-config-standard/blob/master/.eslintrc.json_

## Checklist


.eslintignore
```diff
@@ -0,0 +1,3 @@
+node_modules/
```
townxelliot, line 1: I think this set of ignores is wrong for our project, possibly even redundant.

townxelliot:
Sorry, just realised this is a draft :)

townxelliot (changes requested):
Is this scoped to only lint the application js? I can't see a path anywhere in here which would restrict what gets scanned.

babel.config.js
```diff
@@ -0,0 +1,12 @@
+module.exports = {
+  presets: [
+    [
+      '@babel/preset-env',
+      {
+        targets: {
+          node: 'current',
```
townxelliot, line 7: Why is this here?

townxelliot:
Looks good, just want to be sure we're only scanning application code here. Might also be possible to remove the babel config (not sure if it's being used).

.eslintrc.json
```diff
@@ -0,0 +1,27 @@
...
+    "ecmaVersion": 2022,
+    "sourceType": "module"
+  },
+
+  "parser": "@babel/eslint-parser",
+
+  "env": {
+    "es2021": true,
+    "browser": true,
+    "node": true
```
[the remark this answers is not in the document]
nickdavis2001: OK done

babel.config.js
nickdavis2001: This is required for babel to work

nickdavis2001: OK actually, now changed to not require a config file at all.  you have to explicitly say this in eslintrc

townxelliot (approved):

package.json
```diff
@@ -9,22 +21,17 @@
     "cypress-intellij-reporter": "^0.0.6",
     "cypress-plugin-tab": "^1.0.5",
     "eslint": "^8.23.0",
+    "eslint-config-prettier": "^8.5.0",
```
townxelliot, line 24: You are including prettier, but then not using the plugin at all. We either need to use the prettier rules, or don't require the package. I'd suggest we take it out for now, as all it does is prettifying the code, not finding errors in it. Also, the output from prettier is a bit useless, as every error is categorised as `prettier/prettier`. It also wants us to put semi-colons on the end of every line, which I'd argue is also pointless. So I'd take it out altogether.

package.json
```diff
@@ -1,5 +1,17 @@
...
+  "main": "index.js",
+  "directories": {
+    "doc": "docs",
+    "test": "tests"
+  },
+  "scripts": {
+    "lint:check": "eslint service-front/assets/js cypress/e2e/common"
+  },
   "devDependencies": {
+    "@babel/eslint-parser": "^7.19.1",
```
townxelliot, line 14: Not required, as we don't have any code which needs babel. The default eslint parser is adequate for our purposes.

townxelliot (changes requested):

townxelliot (changes requested):

package.json
nickdavis2001: I stopped using prettier for now because it reported an enormous amount (over 1000) of issues which we'll have to turn off in the config file.  I'll take another look at whether its possible with just a few lines, to turn those off in the config file. I'm not exactly confident short-term that we can get this legacy code to satisfy prettier , due to conflicting requirements, but I guess time will tell.

package.json
```diff
@@ -1,5 +1,17 @@
...
+  "name": "opg-lpa",
+  "description": "The Office of the Public Guardian Lasting Power of Attorney online service: Managed by opg-org-infra &amp; Terraform.",
+  "version": "1.0.0",
+  "main": "index.js",
+  "directories": {
+    "doc": "docs",
+    "test": "tests"
+  },
+  "scripts": {
+    "lint:check": "eslint service-front/assets/js cypress/e2e/common"
```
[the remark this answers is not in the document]
nickdavis2001: Since this reports no further errors with this config, it was a helpful step forward to have the cypress code linted too at this point, as it doesn't interfere with anything else and avoids having to put this back in later and waiting for a further PR. However.... as you wish.... done

package.json
nickdavis2001: Done

.eslintrc.json
```diff
@@ -0,0 +1,27 @@
+{
+  "parserOptions": {
+    "ecmaVersion": 2022,
+    "sourceType": "module",
+    "requireConfigFile": false
+  },
+
+  "parser": "@babel/eslint-parser",
```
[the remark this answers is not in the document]
nickdavis2001: Done

package.json
nickdavis2001: This is now done

.eslintrc.json
```diff
@@ -0,0 +1,27 @@
...
+
+  "env": {
+    "es2021": true,
```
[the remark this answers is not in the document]
nickdavis2001: +1 seems like a good call to downgrade it to es6 given that the code is old and we want to support older browsers

johnnolan (approved):

johnnolan:
@nickdavis2001 looking good! After our conversation this morning you could also move the prettier and eslint config files to service-front/assets and it will be further isolated to that directory ready for when you add the cypress version 
image

nickdavis2001:
+1 @johnnolan good idea , now done :)

johnnolan:
LGTM! :shipit:

johnnolan (approved):

townxelliot:
> +1 @johnnolan good idea , now done :)

Does npm run lint:check still work with the config file moved into a sub-directory?

townxelliot (approved):

PR #1126, opened 2022-09-30
LPAL-992 Add container health checks

ntse:
## Purpose

Add container health checks to the apps for service-front, service-api and service-admin and use the health checks to ensure that at least 2 healthy containers are always running during a deployment.

Fixes LPAL-992

## Approach

For the PHP APIs, a health check script is run every 15 seconds to ensure that PHP-FPM is running and returning data on /health.

If these health checks fail, the containers will not receive traffic.

Deployment_minimum_healthy_percent and deployment_maximum_percent attributes have been added to the admin, API and front ECS service definitions. This will ensure that at least two tasks of each are running and healthy (according to their health checks) at all times. 

The `deployment_minimum_healthy_percent`  is set to 50. This will normally round up to 66% as there are typically 3 tasks running. This ensures that at least 2 out of 3 tasks are healthy at all times, including during deployments,

`deployment_maximum_percent` is set to 200. Normally, this will allow up to 6 tasks to run in order to ensure that at least 2 tasks are always healthy.

## Learning

https://docs.aws.amazon.com/AmazonECS/latest/developerguide/update-service.html
https://www.thatsgeeky.com/2012/02/directly-connecting-to-php-fpm/
https://github.com/ministryofjustice/opg-use-an-lpa/pull/1249

## Checklist

* [] I have added tests to prove my work


service-api/docker/app/health-check.sh
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env sh
+set -e
+
+export SCRIPT_NAME=/health
```
townxelliot, line 4: Presumably these are used by cgi-fcgi? If so, worth a comment to that effect.

service-api/docker/app/Dockerfile
```diff
@@ -5,7 +5,7 @@ RUN adduser -D -g '' appuser
 RUN apk add --upgrade apk-tools
 RUN apk upgrade
 
-RUN apk add --no-cache postgresql-libs icu \
+RUN apk add --no-cache postgresql-libs icu fcgi \
```
townxelliot, line 8: Could we use wget to make the requests? Just wondering why we need fcgi for this.

service-admin/docker/app/health-check.sh
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env sh
```
townxelliot, line 1: Can we have this script in one place and copy it into the two images, as it's identical between them?

docker-compose.yml
```diff
@@ -390,6 +410,16 @@ services:
       PHP_OPCACHE_VALIDATE_TIMESTAMPS: 1
       PHP_IDE_CONFIG: serverName=lpa-admin-app
       OPG_LPA_COMMON_ADMIN_ACCOUNTS: "${OPG_LPA_COMMON_ADMIN_ACCOUNTS}"
+    healthcheck:
+      test:
```
townxelliot, line 414: Is the container testing its own web endpoint here? Is this called by docker behind-the-scenes when other containers want to healthcheck this one?

docker-compose.yml
```diff
@@ -206,6 +206,16 @@ services:
       OPG_LPA_COMMON_REDIS_CACHE_URL: tcp://redis:6379
 
       OPG_LPA_OS_PLACES_HUB_ENDPOINT : "https://api.os.uk/search/places/v1/postcode"
+    healthcheck:
```
townxelliot, line 209: Might be good to just use a script which wraps all this up to reduce verbosity, if possible.

service-admin/docker/app/php-fpm-logging.conf
```diff
@@ -1 +1,2 @@
 access.format = '{"trace_id":"%{HTTP_X_TRACE_ID}e","time_iso8601":"%{%Y-%m-%dT%H:%M:%S%z}T","time_local":"%{%d/%b/%Y:%H:%M:%S %z}T","remote_addr":"%R","real_ip":"%{HTTP_X_REAL_IP}e","real_forwarded_for":"%{HTTP_X_FORWARDED_FOR}e","real_forwarded_proto":"%{HTTP_X_FORWARDED_PROTO}e","request_time":"%d","request_uri":"%{REQUEST_SCHEME}e://%{HTTP_HOST}e%{REQUEST_URI}e","status":"%s","request_method":"%m","http_referrer":"%{HTTP_REFERER}e","http_user_agent":"%{HTTP_USER_AGENT}e","bytes_sent":"%l","http_host":"%{HTTP_HOST}e","server_name":"%{SERVER_NAME}e","server_port":"%{SERVER_PORT}e"}'
+ping.path = '/health'
```
townxelliot, line 2: Is this standard php-fpm config?

terraform/environment/modules/environment/ecs_api.tf
```diff
@@ -297,6 +299,13 @@ locals {
           "protocol" : "tcp"
         }
       ],
+      "healthCheck" : {
+        "command" : ["CMD", "/usr/local/bin/health-check.sh"],
```
townxelliot, line 303: Any reason we can't do the local healthchecks the same way?

townxelliot (changes requested):

service-admin/docker/app/php-fpm-logging.conf
ntse: Yes, this is standard PHP-FPM setting. It's returns a configurable response ("pong!" by default) on the path specified when a FastCGI request comes in.

https://www.php.net/manual/en/install.fpm.configuration.php

docker-compose.yml
ntse: The other containers won't be aware of the health check; the command will run inside of the container that it is targeting. The command according to the below schedule:

      interval: 10s - How often the check is run
      timeout: 15s - If the command takes longer than this to execute, it's considered a failed command.
      retries: 3 - The number of times the command must fail (either timeout or non-zero exit code)
      start_period: 90s - For 90 seconds after the container starts, failed commands won't count towards the retry count unless one of the commands pass then subsequent commands will count towards the retries.
      
If a container is unhealthy (i.e. 3x failed requests) then it will be restarted by the Docker daemon.

service-api/docker/app/Dockerfile
ntse: PHP-FPM will issue a TCP connection reset because wget will try using HTTP instead of the FastCGI protocol. fcgi will exit cleanly (i.e. with exit code 0) if it gets a valid FastCGI response. wget / curl will always exit uncleanly because of the TCP reset.

nickdavis2001 (approved):
🏆

townxelliot (approved):

nickdavis2001 (approved):

PR #1137, opened 2022-10-05
LPAL-1009 Run Cypress tests in parallel

ntse:
## Purpose

Run the Cypress tests in parallel on Github Actions

Fixes LPAL-1009

## Approach

Each Cypress test will be a separate job instead of being a multiple steps of the same job.

## Learning

## Checklist

### DIFF
--- a/.github/workflows/cypress_tests.yml
+++ b/.github/workflows/cypress_tests.yml
@@ -19,11 +19,10 @@ on:
         description: "URL of the frontend"
         required: true
         type: string
-      ephemeral_environment:
-        description: "True if tests are run against ephemeral environment"
-        required: false
-        default: true
-        type: boolean
+      cypress_tags: 
+        description: "Cypress tags to run"
+        required: true
+        type: string
     secrets:
       AWS_ACCESS_KEY_ID_ACTIONS:
         required: true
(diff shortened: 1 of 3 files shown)


.github/workflows/cypress_tests.yml
```diff
@@ -87,131 +86,14 @@ jobs:
...
-          CYPRESS_adminUrl: ${{ inputs.admin_url }}
-          CYPRESS_baseUrl: ${{ inputs.front_url }}
-          CYPRESS_NO_COMMAND_LOG: 1
-          CYPRESS_numTestsKeptInMemory: 1
-          CYPRESS_RUNNER_IN_CI: true
-          CYPRESS_RUNNER_TAGS: "@Signup,@StitchedPF"
-          CYPRESS_RUNNER_BASE_URL: ${{ inputs.front_url }}
-          CYPRESS_RUNNER_ADMIN_URL: ${{ inputs.admin_url }}
-
-        run: |
```
townxelliot, line 101: Have you deliberately removed all of the run definitions? Do you need one of these to be present for anything to happen? Or is this just to test the principle without actually doing anything?

ntse: This may be easier to see here: https://github.com/ministryofjustice/opg-lpa/blob/483b114e9c162f324a921e080cea77c309b853a6/.github/workflows/cypress_tests.yml

There's only one Cypress step now but the entire job gets called multiple times with the different tags

PR #1150, opened 2022-10-10
First check of Cypress Github Actions

ntse:
## Purpose

Fixes LPAL-####

## Approach

## Learning

## Checklist

### DIFF
--- a/.github/workflows/cypress_tests.yml
+++ b/.github/workflows/cypress_tests.yml
@@ -97,6 +97,7 @@ jobs:
           role-session-name: OPGLPACypressTestsCypressS3Monitor
 
       - name: Run Cypress Tests - ${{ inputs.cypress_tags }}
+        uses: cypress-io/github-action@v4.2.0
         env:
           CYPRESS_adminUrl: ${{ inputs.admin_url }}
           CYPRESS_baseUrl: ${{ inputs.front_url }}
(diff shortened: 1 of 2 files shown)


.github/workflows/cypress_tests.yml
```diff
@@ -97,6 +97,7 @@ jobs:
           role-session-name: OPGLPACypressTestsCypressS3Monitor
 
       - name: Run Cypress Tests - ${{ inputs.cypress_tags }}
+        uses: cypress-io/github-action@v4.2.0
```
townxelliot, line 100: This is only going to work for us if it's possible to specify a custom cypress runner and run Python programs in the same runtime. We don't use cypress itself directly: we have a Python script which constructs a cypress command line for us and runs it. The reason for this is so that we can run groups of cypress tags one after the other. In addition, we also run cypress alongside the S3Monitor, which does the work to retrieve emails from S3 and make them available to the cypress test suite.

If there's a way to run a Python script (which runs a node application) within a github task, that might work. But if we want to use the github cypress tasks as they are, we'll probably have to rethink how we structure the tests.

Having said that, the tests marked with @SignupIncluded probably will run directly in cypress, though they will still need the S3Monitor running in parallel to fetch emails from S3.

townxelliot: I just had a quick look at how we currently run the cypress tests, and I think it would be a short step from what we have to just running them natively on the Ubuntu image. It would just mean installing the Python and npm dependencies into that host and calling the cypress runner Python script.

PR #1183, opened 2022-10-27
LPAL-1039 Log duration of slow DB queries

ntse:
## Purpose

Log duration of slow DB queries and disable logging all queries.

Fixes LPAL-1039

## Approach

Use Terraform to adjust the Postgres parameter groups for RDS and Aurora.

## Learning

## Checklist

### DIFF
--- a/terraform/environment/modules/environment/rds.tf
+++ b/terraform/environment/modules/environment/rds.tf
@@ -115,7 +115,7 @@ resource "aws_db_parameter_group" "postgres13-db-params" {
   parameter {
     name         = "log_statement"
     value        = "none"
-    apply_method = "pending-reboot"
+    apply_method = "immediate"
   }
 
 
--- a/terraform/environment/modules/environment/rds.tf
+++ b/terraform/environment/modules/environment/rds.tf
@@ -114,7 +114,7 @@ resource "aws_db_parameter_group" "postgres13-db-params" {
 
   parameter {
     name         = "log_statement"
-    value        = "all"
+    value        = "none"
     apply_method = "pending-reboot"
   }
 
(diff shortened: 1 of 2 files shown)


terraform/environment/modules/environment/rds.tf
```diff
@@ -124,6 +124,12 @@ resource "aws_db_parameter_group" "postgres13-db-params" {
     value        = "1440"
     apply_method = "immediate"
   }
+
+  parameter {
+    name         = "log_duration"
```
townxelliot, line 129: Could you please give me a brief description of what effect `log_duration` has? What does "1" signify?

terraform/environment/modules/environment/rds.tf
```diff
@@ -114,8 +114,8 @@ resource "aws_db_parameter_group" "postgres13-db-params" {
 
   parameter {
     name         = "log_statement"
```
townxelliot, line 116: How does this change behaviour please?

townxelliot (changes requested):

terraform/environment/modules/environment/rds.tf
ntse: Log duration will print the time taken for an SQL query to execute next to the statement. 1 equates to True

terraform/environment/modules/environment/rds.tf
ntse: We will stop logging all queries and only log queries that take 500ms > which, according to Performance Insights, are rare

nickdavis2001 (approved):

townxelliot (approved):

PR #1186, opened 2022-10-31
LPAL-467 Add psalm scan to CI build and release number of PHP linters in pre-commit hooks

townxelliot:
## Purpose

[LPAL-467](https://opgtransform.atlassian.net/browse/LPAL-467)

## Approach

See ticket.

## Learning

n/a

## Checklist


service-front/psalm.xml
```diff
@@ -8,7 +8,6 @@
 >
     <projectFiles>
         <directory name="module/Application/src" />
-        <directory name="../shared/logging/module/MakeLogger/src/" />
```
nickdavis2001, line 11: Do you not wish to carry on scanning this?  Or has it moved or something?

.pre-commit-config.yaml
```diff
@@ -18,16 +18,9 @@ repos:
     rev: 1.4.0
     hooks:
     - id: php-lint-all
-    - id: php-cs
```
nickdavis2001, line 21: if removal of this because it would duplicate psalm?

service-front/psalm.xml
townxelliot: psalm can find it anyway as it uses the generated autoload.php file, so this is redundant.

.pre-commit-config.yaml
townxelliot: phpcbf actually uses php-cs as the underlying scanner. So in effect, we're doing the same scan twice.

townxelliot: I removed phpstan as it seems pointless to have two linters doing the same job, which may in fact contradict each other. phpcbf does a slightly different job, as it enforces code style, rather than code correctness.

service-front/psalm.xml
townxelliot: Actually, though, I think you're right: one of the configurations should scan that directory. Or maybe the shared directory needs its own scan.

nickdavis2001 (approved):

service-front/module/Application/src/Model/Service/Signatures/DateCheck.php
```diff
@@ -146,13 +146,8 @@ public static function checkDates(
 
                 // Donor must be first
                 if ($donor > $timestamp) {
-                    $message = "${donorText} must be the first person to sign the LPA. ";
```
townxelliot, line 149: $isDraft is always false, otherwise you're not in this block at all; therefore the `if ($isDraft)` condition never passes, and the code inside is dead.

service-front/module/Application/tests/Form/Lpa/TrustCorporationFormTest.php
```diff
@@ -68,11 +68,9 @@ public function testValidateByModelInvalid()
         $this->assertEquals([
             'name' => [
                 0 => 'cannot-be-blank',
-                1 => 'must-be-greater-than-or-equal:1'
```
townxelliot, line 71: Changes to opg-lpa-datamodels mean that this test is now incorrect. This was not picked up previously as unit tests weren't running.

PR #1187, opened 2022-11-01
LPAL-797 Mend the cypress tests so they can all run standalone

townxelliot:
## Purpose

Fixes LPAL-####

## Approach

## Learning

## Checklist


cypress/e2e/ReplacementAttorneysHW.feature
```diff
@@ -23,7 +23,7 @@ Feature: Add Replacement Attorneys to a Health and Welfare LPA
         And I can find "postcode-lookup"
         And I can find "name-title" with 8 options
         When I select "Ms" on "name-title"
-        And I force fill out
+        And I quickly fill out
```
nickdavis2001, line 26: What does 'quickly' really mean here?  wonder if this could be named a bit more precisely?

cypress/e2e/DateCheck.feature
```diff
@@ -17,32 +19,27 @@ Feature: Check signature dates
...
     And I can see a reminder to sign continuation sheet 4
     And I can see that the donor "Mr Christopher Robin" cannot sign
 
   Scenario: Displays the correct references to donor/applicant in errors when they cannot sign
-    Given I visit the dashboard
-    When I click "check-signing-dates" for LPA ID 26997335999
-    And I click element marked "Check dates"
-    And the visually-hidden legend for "date-check-donor-date" states "Check signature dates for the person who signed on behalf of the donor"
-    Then I can see validation errors refer to the person signing on behalf of the donor, who is also the applicant
+    Given I visit "/lpa/26997335999/date-check"
```
nickdavis2001, line 30: The specific url here bothers me somewhat.  The way it was before , didn't specify the exact url, so that could move. This way we are tied to a specific path.  I suppose its unlikely to change, but still.   Also the hard-coded LPA ID bothers me too a bit. But I suppose there has to be some way of identifying the exact test fixture we want to look at

cypress/e2e/ReplacementAttorneysHW.feature
townxelliot: I added another method to fill a form in which doesn't use `type()`. This just speeds up filling forms. It can't be used everywhere as it doesn't trigger change events the same way (so can't be used to test the client-side validation).

cypress/e2e/DateCheck.feature
townxelliot: The LPA ID has to be there to get the fixture we want, as you say. As for the path: the alternative is to search for the LPA ID, then click on the link. That would remove the hard-coded path, but would introduce about 3 extra steps for each test case (load the dashboard, enter text in the search box, press search, click on the link). We could do that if you think it's important. On the other hand, these paths are unlikely to change; and if they do, the tests will break and we can fix them.

nickdavis2001: OK

nickdavis2001 (dismissed):

cypress/e2e/DateCheck.feature
townxelliot: Would you prefer if the path was hidden behind some plain English?

cypress/e2e/ReplacementAttorneysHW.feature
townxelliot: If you've got a suggestion for a better name, I'm happy to change it.

cypress/e2e/DateCheck.feature
nickdavis2001: Actually that might be a good thing, yes. :) It just hides the implementation so that in theory if the link moved or something, tests wouldn't need to change.

cypress/e2e/ReplacementAttorneysHW.feature
nickdavis2001: Or alternatively, should we make everything do it the "quick" way and take out the word "quickly" ?  At the moment some of the filling out is done this way and some the conventional way which is inconsistent and could makes tests behave weirdly maybe?  But this could be another story and PR of course.  :). What do you think?

townxelliot: The problem is that you can't use the "quickly" technique for all of the tests. Where the user interface reacts to changes in an input (e.g. where we prompt the user about how old/young the attorney is when they edit their birthdate), the "quickly" method doesn't work: in effect, it just sets the value attribute on the input, which doesn't trigger the change events on the input. So we can either keep the "quickly" approach (i.e. fill form inputs en masse where the JS event triggering doesn't matter), which as you rightly say is inconsistent; or revert to using the "force" approach everywhere (which potentially slows things down).

cypress/e2e/DateCheck.feature
townxelliot: I'll do that.

nickdavis2001 (approved):

PR #1190, opened 2022-11-02
LPAL-1046 Replace dockerignore files

ntse:
## Purpose

Use Dockerignore file to reduce build context size.

Fixes LPAL-1046

## Approach

Remove the unnecessary / unused dockerignore files in each individual app's directory and replace with dockerignore file in the root directory.

## Learning

## Checklist


.dockerignore
```diff
@@ -0,0 +1,9 @@
+.git
+.gitignore
+.github
+terraform
+docs
+
+*/LICENSE
+*/LICENSE.md
+*/README.md
```
townxelliot, line 9: If this is case-sensitive, bear in mind that one of our repos irritatingly has a readme.md (lowercase).

townxelliot: I would probably also add `*/build`, `.phpunit.result.cache`, `Makefile`, `*/Makefile` (don't think we run `make` in any docker builds), `CODEOWNERS`, `.idea`, `.php-cs-fixer.cache`, `metadata.json`, `tests/sirius-gateway-mock`, `tests/load` (I think we shouldn't ignore `tests/python-api-client` as the cypress container needs it, though I also think that should be under the cypress directory for that reason). Also make it `LICENSE*` and `*/LICENSE*` to cover the variety of licence files in different directories. (It's a shame the `**/filename` syntax isn't supported like in .gitignore.)

townxelliot (changes requested):

townxelliot (approved):

PR #1197, opened 2022-11-07
LPAL-1038 fix js issue in feedback form

nickdavis2001:
## Purpose

_Bugfix for feedback form_

Fixes LPAL-1038

## Approach

_Fix logic in js and add extra cypress testing. Tidy the twig template_

## Checklist


cypress/e2e/Feedback.feature
```diff
@@ -19,7 +19,16 @@ Feature: Feedback
             | Select a rating for this service |
             | Do not forget to leave your feedback in the box |
         And I see "Error" in the title
-        When I select satisfied
+        # click through all the radio buttons due to history of bug in this area (see LPAL-1038)
```
townxelliot, line 22: We probably need to do something more than this to prove that the radio button is really being selected. (Clicks on the radio buttons do work, it's just that the value isn't set correctly.) Could be as simple as checking that input on the form; or could intercept the submitted form and check the value of the radio there. The latter would be more complicated but probably more indicative that the form is working properly.

townxelliot (changes requested):

cypress/e2e/Feedback.feature
nickdavis2001: Done

townxelliot: I was expecting some extra assertions that the `rating` input on the form has the expected value. Am I missing something?

nickdavis2001: Whoops. Had committed on wrong git branch. 🤦  Should be sorted now

townxelliot (approved):

PR #1201, opened 2022-11-09
LPAL-972  shared folder reorg

nickdavis2001:
## Purpose

_Re-organise the shared folder to make it easier to add other shared modules_

## Approach

_Change structure of shared folder and update references to Logger in php, to look in the right place now its moved. Add shared folder to psalm scan_

## Checklist


.github/workflows/analysis-psalm.yml
```diff
@@ -26,6 +26,8 @@ jobs:
             path: "./service-admin"
           - name: service-pdf
             path: "./service-pdf"
+          - name: shared
```
townxelliot, line 29: You're probably going to need an autoloader here, and possibly a composer.json file, as this has dependencies of its own (if you're psalm scanning it).

service-api/module/Application/tests/Bootstrap.php
```diff
@@ -14,18 +14,18 @@
     $baseDirs = [
         __DIR__ . '/',
         __DIR__ . '/../src/',
-        __DIR__ . '/../../../../shared/logging/module/MakeLogger/src/',
+        __DIR__ . '/../../../../shared//module/MakeShared/src/',
```
townxelliot, line 17: Extra forward-slash here.

shared/composer.json
```diff
@@ -0,0 +1,26 @@
+{
```
townxelliot, line 1: Ah, I see you added a composer.json :)

shared/psalm.xml
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0"?>
```
townxelliot, line 1: And a psalm.xml :) Looks like you probably just need to add shared/ to the run-composers target in the Makefile so the vendor folder is populated. It might also be necessary to run composer install in the CI build; and possibly we need the composer.lock in the shared folder (atm it's not checked in).

townxelliot (approved):

shared/psalm.xml
nickdavis2001: All done :)

townxelliot (dismissed):

townxelliot (dismissed):

PR #1208, opened 2022-11-14
LPAL-1053 Remove use of fpassthru in PHP code

townxelliot:
## Purpose

[LPAL-1053]()

## Approach

See ticket.

## Learning

n/a

## Checklist


cypress/e2e/Admin.feature
```diff
@@ -142,3 +132,6 @@ Feature: Admin
         | test-no-email-no-number |
         | test-no-num             |
         | test                    |
+
+    # LPAL-1053: check feedback export to CSV works correctly
+    And I can export feedback and download it as a CSV file
```
nickdavis2001, line 137: Given that the way of generating the csv has been completely changed (which is good :) ). should there be a test that ensures the contents of the csv are OK?  Assuming we didn't have one before...  which we should have of course.  Might be better as unit test?

townxelliot: There are very few unit tests on any of this code. I could write a test to cover this relatively easily, I think.

townxelliot: Not easy to unit test as it manipulates headers and writes directly to the output stream. I tried a work-around which is meant to work for this, but couldn't get it to work. I'll extend the cypress test instead.

townxelliot: @nickdavis2001 I have added some additional checks of the structure of the data, if you could please re-review. Doing this in a unit test proved tricky, as testing handlers (with their csrf tokens etc.) was unwieldy and unintuitive.

nickdavis2001 (approved):

PR #1209, opened 2022-11-14
LPAL-1045 Combine opg-data-models repo into main Make repo

nickdavis2001:
## Purpose

_LPAL-1045 Combine opg-data-models repo into opg-lpa to ease maintenance of code base_

## Approach

_Bring DataModels and associated tests into the shared folder in opg-lpa. Add necessary config files to run psalm. Fix all psalm errors. Modify front, api ,pdf , admin services to point at this new location _

## Checklist


service-admin/composer.json
```diff
@@ -67,7 +61,9 @@
         "phpunit/phpunit": "^9.5.10",
         "roave/security-advisories": "dev-master",
         "squizlabs/php_codesniffer": "^3.6.1",
-        "vimeo/psalm": "^4.23"
+        "vimeo/psalm": "^4.23",
```
townxelliot, line 64: Are the following two packages in the dev section? If they are, they should be moved into the main (runtime) section as I believe they are runtime dependencies.

service-front-v2/app/templates/component.html
```diff
@@ -0,0 +1,141 @@
+{% extends "base.html" %}
```
townxelliot, line 1: This looks like it's been committed in error.

service-front/module/Application/tests/Controller/Authenticated/Lpa/StatusControllerTest.php
```diff
@@ -46,13 +47,14 @@ public function testIndexActionWithReturnUnpaid()
     }
 
     public function testIndexActionInvalidStatus()
-    {/** @var StatusController $controller */
+    {
+/** @var StatusController $controller */
```
townxelliot, line 51: Line 51 needs indenting.

townxelliot (approved):
A couple of minor things, but nice work, it was a tough one! I didn't review the individual opg-lpa-datamodels files you copied over, as I assume you only made syntactic and namespace changes to them.

townxelliot (changes requested):
A few minor comments, but well done for sticking at it, it was a tough one! I didn't review opg-lpa-datamodels as I assumed any changes made there would be syntax and namespace related. (Though I will go and look at that one function we worked on and the unit tests.)

shared/module/MakeShared/tests/Bootstrap.php
```diff
@@ -0,0 +1,5 @@
+<?php
```
townxelliot, line 1: I think this will deal with loading files, so we shouldn't need the autoload.php file. I would try removing the latter and see if everything still works.

shared/module/MakeShared/tests/autoload.php
```diff
@@ -0,0 +1,40 @@
+<?php
```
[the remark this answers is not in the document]
townxelliot: I've just confirmed locally, this can be removed.

service-admin/composer.json
```diff
@@ -67,7 +61,9 @@
...
+        "symfony/validator": "^4.2",
+        "egulias/email-validator": "~1.2"
```
townxelliot, line 66: I thought laminas did the email validation? It probably could (but that might be for another ticket).

townxelliot (changes requested):

service-front-v2/app/templates/component.html
nickdavis2001: Agreed. now fixed.

townxelliot:
One broader comment for another ticket: rather than including the dependencies for shared/ in every composer.json, we should probably do a separate composer install using shared/composer.json, but put the packages into the app/vendor directory. That way you can list the shared/ deps in one place, but we install shared's dependencies into the same location as service-front's (or service-api's etc.) dependencies.

townxelliot:
shared/composer.json needs a new description which better explains what's in this directory (I can't comment on it inline because it hasn't changed).

shared/module/MakeShared/tests/autoload.php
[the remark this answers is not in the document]
nickdavis2001: OK now removed, and .gitignored too

shared/module/MakeShared/tests/Bootstrap.php
nickdavis2001: Done

service-admin/composer.json
nickdavis2001: These packages turn out not to be needed in either section, so now removed

service-admin/composer.json
nickdavis2001: OK removed because turned out not needed

nickdavis2001:
> shared/composer.json needs a new description which better explains what's in this directory (I can't comment on it inline because it hasn't changed).

Done :)

townxelliot (dismissed):

townxelliot (approved):

PR #1219, opened 2022-11-23
LPAL-1064 remove old files that are no longer relevant

nickdavis2001:
## Purpose

_For improved security and ease of maintenance, remove old legacy files that are no longer relevant_

## Approach

_Targeted removal of files and check that everything is still working. There is an assortment of files. Includes a number in service-api that would only be used if this is front-end, but never would because its just a back-end rest api. Note - the API still cleanly returns a 404 if a call is made to a non-existing path from the python test api, or with curl_

## Checklist


townxelliot (dismissed):

townxelliot:
What happens if you request a path on the API which doesn't exist? Or if an internal server error happens? Do we get JSON back?

townxelliot:
testCoverage.sh can also be pulled.

townxelliot:
Once you've merged the changes phpunit running outside docker, dockerignore will need some changes for the files you've removed (like testCoverage.sh).

townxelliot:
When you reinstate the HTML for the API endpoints which otherwise show an error, I suggest you reduce them as much as possible so we only respond with an HTML file. So any references to JS and CSS can be stripped, along with the corresponding files.

townxelliot (approved):

PR #1224, opened 2022-11-25
LPAL-1062 WAF "on" false positive workaround

townxelliot:
## Purpose

Fixes LPAL-####

## Approach

## Learning

## Checklist

### DIFF
--- a/service-front/assets/js/moj/moj.modules/moj.form-popup.js
+++ b/service-front/assets/js/moj/moj.modules/moj.form-popup.js
@@ -117,6 +117,44 @@
       });
     },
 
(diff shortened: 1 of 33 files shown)


townxelliot:
This solution is not sufficient, as we still get the error from forms which don't use Ajax; i.e. it's not that we are using form-encoded data in our Ajax request bodies, but that the WAF rules create false positives for both JS and non-JS form submissions.

[LPAL-1074](https://opgtransform.atlassian.net/browse/LPAL-1074) is the follow-up issue to apply the necessary change to the WAF.

PR #1227, opened 2022-11-29
LPAL-1074 Allow ON keyword in JSON body

ntse:
## Purpose

Allow the ON keyword in JSON body requests

Fixes LPAL-1074

## Approach

Add an exception for requests matching the request .*[oO][nN].*=.*$

## Learning

https://support.console.aws.amazon.com/support/home#/case/?displayId=11374935181

## Checklist


nickdavis2001 (dismissed):

terraform/region/modules/region/waf.tf
```diff
@@ -84,6 +123,18 @@ resource "aws_wafv2_web_acl" "main" {
...
 }
 
+// Avoid false-positives for the AWSManagedRulesCommonRuleSet rule - see AWS support ticket 11374935181
+resource "aws_wafv2_regex_pattern_set" "allow_on_keyword" {
+  name        = "on-keyword"
+  description = "On keyword regex pattern set"
+  scope       = "REGIONAL"
+
+  regular_expression {
+    regex_string = ".*[oO][nN].*=.*$"
```
townxelliot, line 133: This will match for multipart form encoded POSTs, which is fine as we're no longer bothering with converting the popup Ajax requests to JSON.

townxelliot:
Note that these rules cover standard multipart form encoded POST requests, rather than JSON requests.

townxelliot:
Is there any way we can test this before it goes to live? I wonder if we should do the ticket to enable WAF in dev first?

ntse:
This isn't possible until the AWS provider version is updated to 4.29.0. This is not yet released.

https://github.com/hashicorp/terraform-provider-aws/blob/v4.29.0/CHANGELOG.md

PR #1258, opened 2022-12-12
LPAL-225 move more shared components to shared area

nickdavis2001:
## Purpose

_LPAL-225 move more shared components to shared area _

## Approach

_Move files and change paths in Dockerfiles. Note that for service-pdf the logging conf is removed entirely, as it turns out not to be used. Where files are not the same as in other services, they have .service-admin or service-pdf on the end for now. We may be able to rationalise these in future_

## Learning

## Checklist


shared/docker/app/app-php.ini
```diff
@@ -0,0 +1,10 @@
+log_errors = On
```
townxelliot, line 1: Is there a reason we can't use this for service-pdf? Does the opcache stuff mean it's unsuitable for a non-web app?

shared/docker/app/php-fpm-logging.conf
```diff
@@ -1,2 +1,2 @@
 access.format = '{"trace_id":"%{HTTP_X_TRACE_ID}e","time_iso8601":"%{%Y-%m-%dT%H:%M:%S%z}T","time_local":"%{%d/%b/%Y:%H:%M:%S %z}T","remote_addr":"%R","real_ip":"%{HTTP_X_REAL_IP}e","real_forwarded_for":"%{HTTP_X_FORWARDED_FOR}e","real_forwarded_proto":"%{HTTP_X_FORWARDED_PROTO}e","request_time":"%d","request_uri":"%{REQUEST_SCHEME}e://%{HTTP_HOST}e%{REQUEST_URI}e","status":"%s","request_method":"%m","http_referrer":"%{HTTP_REFERER}e","http_user_agent":"%{HTTP_USER_AGENT}e","bytes_sent":"%l","http_host":"%{HTTP_HOST}e","server_name":"%{SERVER_NAME}e","server_port":"%{SERVER_PORT}e"}'
```
townxelliot, line 1: I'd like to change the name of this file. At some point, it also took on config for the ping.path, which it didn't originally have in it, meaning it's no longer just php-fpm-logging config. Would it make sense to rename this to something more generic? Alternatively, keep the logging config in one file and add another file for ping config?

townxelliot (changes requested):
Eventually, we shouldn't need the .service-admin suffixes for some of these files, as I'm hoping the config is generic across all the web services. Is it worth removing the suffix for service-admin, while retaining service-pdf suffixes (which do need to remain separate)?

shared/docker/app/php-fpm-logging.conf
nickdavis2001: OK, now renamed to more generic app-php.conf. (to be in line with app-php.ini). We could in future seperate ping config and logging config although that might add complexity. Seems harmless to have various config in 1 file now it has generic name?

shared/docker/app/app-php.ini
nickdavis2001: OK, by using opcache in pdf I have now made them use the same file

shared/docker/app/php-fpm-logging.conf
nickdavis2001: Now renamed back as this turns out to be a reserved name and renaming it would break everything.

townxelliot (approved):

PR #1278, opened 2022-12-20
LPAL-970 Run Cypress tests on Github Actions runner

ntse:
## Purpose

Reduce the amount of time taken for Cypress tests to run

Fixes LPAL-970

## Approach

Run Cypress tests directly on Github Actions runner, removing the need to build the Cypress docker container in CI and remove the need to docker pull it.

## Learning

## Checklist

[LPAL-970]: https://opgtransform.atlassian.net/browse/LPAL-970?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/workflows/cypress_tests.yml
+++ b/.github/workflows/cypress_tests.yml
@@ -79,6 +79,7 @@ jobs:
       - uses: actions/cache@v2
         with:
           path: |
+            ~/.cache/Cypress
             node_modules
           key: node-cache-${{ runner.os }}-${{ hashFiles('package-lock.json') }}
 
--- a/.github/workflows/cypress_tests.yml
+++ b/.github/workflows/cypress_tests.yml
@@ -96,6 +96,10 @@ jobs:
           role-duration-seconds: 3600
           role-session-name: OPGLPACypressTestsCypressS3Monitor
 
+      - uses: actions/setup-node@8c91899e586c5b171469028077307d293428b516 # pin@v3.5.1
+        with:
+          node-version: 16
+
       - name: Run Cypress Tests - ${{ inputs.cypress_tags }}
         env:
           CYPRESS_adminUrl: ${{ inputs.admin_url }}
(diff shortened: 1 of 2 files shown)


.github/workflows/cypress_tests.yml
```diff
@@ -96,6 +75,21 @@ jobs:
...
+          path: |
+            node_modules
+          key: node-cache-${{ runner.os }}-${{ hashFiles('package-lock.json') }}
+
+      - uses: actions/setup-node@8c91899e586c5b171469028077307d293428b516 # pin@v3.5.1
+        with:
+          node-version: 16
+
+      - name: Cypress install
+        uses: cypress-io/github-action@v5.0.2
```
townxelliot, line 89: Does this fix us to a specific version of cypress, or is this just so we get all the necessary dependencies installed?

townxelliot (changes requested):

townxelliot:
@ntse looks good. Does this lock the version of node to a known-good one? I'm not sure if this is enforceable somewhere, perhaps in package.json? As we're now not using the cypress images, we might want to ensure this is the case.

We can also remove the cypress/Dockerfile if we are now running directly inside a github action.

.github/workflows/cypress_tests.yml
ntse: I've removed this now so we're just using the version of Cypress in the package.json file

ntse:
> @ntse looks good. Does this lock the version of node to a known-good one? I'm not sure if this is enforceable somewhere, perhaps in package.json? As we're now not using the cypress images, we might want to ensure this is the case.
> 
> We can also remove the cypress/Dockerfile if we are now running directly inside a github action.

I think we'll still need the cypress/Dockerfile for running Cypress locally with `make cypress-local`

townxelliot (approved):

PR #1314, opened 2023-01-19
LPAL-1080 Upgrade service-admin to PHP 8.2

nickdavis2001:
## Purpose

_Upgrade service-admin to PHP _

## Approach

_Update versions in composer.json, have CI use 8.2 version of phpunit and psalm_

## Checklist


townxelliot (changes requested):
You'll also need to modify the github action config so we use php8.2 for unit tests and psalm scan.

service-admin/composer.json
```diff
@@ -13,13 +13,13 @@
...
-        "php": "^8.1",
+        "php": "^8.2",
         "aws/aws-sdk-php": "^3.81",
         "firebase/php-jwt": "^6.0",
         "guzzlehttp/guzzle": "^6.3",
         "laminas/laminas-authentication": "^2.8",
-        "laminas/laminas-cache": "^3.1.2",
-        "laminas/laminas-cache-storage-adapter-memory": "2.0",
+        "laminas/laminas-cache": "^3.9",
+        "laminas/laminas-cache-storage-adapter-memory": "2.2",
```
townxelliot, line 22: I would make this "^2.2"

townxelliot (changes requested):

service-admin/composer.json
```diff
@@ -13,13 +13,13 @@
         }
     },
     "require": {
-        "php": "^8.1",
+        "php": "^8.2",
```
townxelliot, line 16: I would probably fix this to "8.2". That way we can't accidentally move it to 8.3 without realising.

townxelliot (changes requested):

service-admin/composer.json
nickdavis2001: done

service-admin/composer.json
nickdavis2001: done

townxelliot:
Because we had to revert the PHP 8.2 migration for service-front, it broke your .github/workflows changes. I think we should probably attempt to push service-admin to PHP 8.2 as it's low risk and shouldn't suffer any performance drops as it gets hardly any load.

nickdavis2001:
@townxelliot I have fixed the conflicts :)

townxelliot (approved):

PR #1344, opened 2023-02-23
LPAL-378 - Add account deletion test

mabely:
## Purpose

Add account deletion integration tests for creating and deleting a user from service-front, and checking that it is reflected on service-admin.
[LPAL-378](https://opgtransform.atlassian.net/browse/LPAL-378)

## Approach

## Learning

## Checklist

[LPAL-378]: https://opgtransform.atlassian.net/browse/LPAL-378?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/actions/detect_changes/action.yml
+++ b/.github/actions/detect_changes/action.yml
@@ -50,7 +50,7 @@ runs:
         id: determine_changes
         shell: bash
         run: |
-          if [[ "${{ steps.detect_changes.outputs.changes }}" == "true"  || "${{ steps.detect_shared_changes.outputs.changes }}" == "true" || "${{ steps.detect_seeding_changes.changes }}" == "true" ]]; then
+          if [[ "${{ steps.detect_changes.outputs.changes }}" == "true"  || "${{ steps.detect_shared_changes.outputs.changes }}" == "true" || "${{ steps.detect_seeding_changes.outputs.changes }}" == "true" ]]; then
             echo "changes=true" >> $GITHUB_OUTPUT
           else
             echo "changes=false" >> $GITHUB_OUTPUT
--- a/.github/actions/detect_changes/action.yml
+++ b/.github/actions/detect_changes/action.yml
@@ -18,7 +18,7 @@ runs:
         run: |
           # Fetch the git history for the branch. Do not error if we already have the history
           git fetch --unshallow || true
-  
+
       - name: Detect changes in ${{ inputs.path }} to main using git diff
         id: detect_changes
         if: github.ref != 'refs/heads/main'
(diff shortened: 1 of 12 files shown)


townxelliot:
Failing because of this @mabely:

```  1) AdminUserAccount
       Delete user from user-facing site:
     CypressError: Timed out retrying after 12000ms: The command was expected to run against origin `https://localhost:7003` but the application is at origin ```

Presumably you are still inside a cy.origin() block but trying to access pages on service-front?

**Comment by mabely** [comment]
@townxelliot I suspect it's because I need to reference adminUrl as the cypress env var, initially it didn't work locally possibly because I was on a very old version of Cypress so I hard coded it to localhost. I've fixed that now, will need to see if it fixes in CI

File: .github/actions/detect_changes/action.yml [hunk 1]
```diff
@@ -37,12 +37,21 @@ runs:
...
+          git diff --quiet HEAD origin/main -- scripts/non_live_seeding && \
+          echo "changes=false" >> $GITHUB_OUTPUT || \
+          echo "changes=true" >> $GITHUB_OUTPUT
+
       - name: Determine if changes were detected
         id: determine_changes
         shell: bash
         run: |
-          if [[ "${{ steps.detect_changes.outputs.changes }}" == "true"  || "${{ steps.detect_shared_changes.outputs.changes }}" == "true" ]]; then
+          if [[ "${{ steps.detect_changes.outputs.changes }}" == "true"  || "${{ steps.detect_shared_changes.outputs.changes }}" == "true" || "${{ steps.detect_seeding_changes.changes }}" == "true" ]]; then
```
townxelliot, line 53: does steps.detect_seeding_changes.changes need to be steps.detect_seeding_changes.outputs.changes?

townxelliot: An alternative approach: because of how git diff works, I think we could also add an additional directory to inputs.path, e.g. in .github/workflows/docker_job.yml:

```
          - image_name: online-lpa/seeding_app
            dockerfile_path: ./service-seeding/docker/app
            service_path: ./service-seeding ./scripts/non_live_seeding
```

Even better, we could just move the scripts into service-seeding, as they aren't used for anything else afaik.

townxelliot: I would probably just leave this who you've change it for now, though I think you might possibly need outputs as explained above.

service-admin/src/App/templates/app/user-search.phtml
```diff
@@ -54,63 +54,67 @@ $this->addErrorMap([
...
-
-                    <?php if (array_key_exists('isActive', $user)) { ?>
-                        <span><?= ($user['isActive'] ? 'Activated' : 'Not activated'); ?></span>
-                    <?php } elseif (array_key_exists('isDeleted', $user) && $user['isDeleted']) { ?>
-                        <span>Deleted</span>
-                    <?php } ?>
-                </li>
-
-                <?php if (isset($user['activatedAt'])) { ?>
+                <div class="user-search-result">
```
townxelliot, line 57: A `<div>` shouldn't be nested inside a `<ul>`. Is there a reason this markup needs to change?

townxelliot (changes requested):

.github/actions/detect_changes/action.yml
mabely: Ah thanks, yes it is supposed to be, I missed it somehow.

Moving the seeding scripts to service-seeding would be quite good actually, there's nothing else in that directory.

service-admin/src/App/templates/app/user-search.phtml
mabely: I didn't know that was a rule, I originally did it for the cypress tests and to conform to user-find, but I then should have had the `<ul>` nested inside the `<div>` in that case. I'll change it to that unless there's preference for having the class on the `<ul>`?

service-front/package-lock.json
```diff
@@ -10,209 +10,356 @@
       "hasInstallScript": true,
       "license": "MIT",
       "dependencies": {
+        "grunt-contrib-watch": "^1.1.0",
```
nickdavis2001, line 13: Is this file actually meant to be commited? I think you might have the same problem as me, that package.json keeps getting regenerated locally.  @townxelliot has made another PR to stop that happening, so he may comment here.

mabely: It was meant to be but it was a decision I regret, I didn't think the branch would end up hitting stumbling blocks and be delayed. At the time it seemed like we should update the lock file to keep it up to date because it would keep regenerating locally as well like yours, but this PR dragged on for so long that it was a relic from before @townxelliot fixed it. To fix the conflicts, I removed the lock file, re-ran `npm install .` then committed the file it generated instead

nickdavis2001: 👍

townxelliot (approved):

PR #1384, opened 2023-03-10
LPAL-1170 Prevent unwanted updates to service-front/package-lock.json

townxelliot:
## Purpose

[LPAL-1170](https://opgtransform.atlassian.net/browse/LPAL-1170)

## Approach

see ticket

## Learning

nvmrc et al

## Checklist

[LPAL-1170]: https://opgtransform.atlassian.net/browse/LPAL-1170?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.github/workflows/docker_job.yml
```diff
@@ -113,11 +113,11 @@ jobs:
             echo "build=false" >> $GITHUB_OUTPUT
           fi
 
-      - name: Use Node.js 16.x
+      - name: Use Node.js 19.x
         uses: actions/setup-node@64ed1c7eab4cce3362f8c340dee64e5eaeef8f7c # tag=v3.6.0
         if: ${{ (matrix.image_name == 'online-lpa/front_web') && steps.build_image.outputs.build == 'true' }}
         with:
-          node-version: 16.x
+          node-version: 19.x
```
johnnolan, line 120: Not sure if its possible here but if you have a .nvmrc file like here https://github.com/ministryofjustice/opg-modernising-lpa-docs/blob/main/.nvmrc you can reference that file like so https://github.com/ministryofjustice/opg-modernising-lpa-docs/blob/main/.github/workflows/pr_workflow.yml#L19 to always use the version that gets bumped and used in the rest of the code base

townxelliot: > Not sure if its possible here but if you have a .nvmrc file like here https://github.com/ministryofjustice/opg-modernising-lpa-docs/blob/main/.nvmrc you can reference that file like so https://github.com/ministryofjustice/opg-modernising-lpa-docs/blob/main/.github/workflows/pr_workflow.yml#L19 to always use the version that gets bumped and used in the rest of the code base

I'll do that. Can you do the same with .npmrc?

townxelliot: @johnnolan see what you think; I've left the npmrc file in service-front, as that's where it really matters: we want to enforce the engines specified in package.json for node-build-assets (which uses service-front as its base directory). I've put .nvmrc in the root directory as it is used in several build steps.

johnnolan (dismissed):

johnnolan (approved):

PR #1415, opened 2023-03-13
LPAL-1168 remove service-front-v2

nickdavis2001:
## Purpose

_LPAL-1168 Remove service-front-v2 now that we know Make an LPA site has limited lifespan due to being superseded by Modernise, and the possibility of parts of Make being ported to python is no longer there_

## Approach

_Take out service-front-v2 and associated machinery and cypress test_

## Checklist


service-front/docker/web/etc/confd/templates/default.conf
```diff
@@ -126,24 +122,7 @@ server {
     #    internal;
     }
 
-    location = /favicon.ico {
```
townxelliot, line 129: Do we need to keep this line for the favicon?

townxelliot (changes requested):

service-front/docker/web/etc/confd/templates/default.conf
nickdavis2001: Good catch yes I think we still need it so added back

townxelliot:
You will also need to remove the feedbackdb github tasks, as these are no longer required.

PR #1418, opened 2023-03-13
LPAL-1166 upgrade to notifications-php-client v5

nickdavis2001:
## Purpose

_LPAL-1166 upgrade to notifications-php-client v5 and associated package upgrades_

## Approach

_Updates to composer.json, mods to php code_

## Learning

## Checklist


townxelliot (approved):
I suggest we get some UAT on this, as it touches the emails.

townxelliot (approved):

service-api/composer.json
```diff
@@ -7,7 +7,7 @@
     "repositories": [
     ],
     "require": {
-        "php": ">=8.1 <8.2.0",
+        "php": ">=8.1 <8.2.2",
```
townxelliot, line 10: We don't want any 8.2 PHP versions in composer.json files, except for in service-admin. This causes a potential issue if a PHP library is compatible with 8.2 but not 8.1.

service-front/composer.json
```diff
@@ -10,11 +10,12 @@
     ],
     "homepage": "https://github.com/ministryofjustice/opg-lpa",
     "require": {
-        "php": ">=8.1 <8.2.0",
-        "guzzlehttp/guzzle": "^6.3",
+        "php": "^8.1",
+        "php": ">=8.1 <8.2.2",
```
townxelliot, line 14: ditto, we need versions of PHP < 8.2.0

townxelliot (dismissed):

service-front/composer.json
townxelliot, line 14: ditto, we only want PHP < 8.2.0 for service-front for now

nickdavis2001: OK sorted

service-api/composer.json
nickdavis2001: OK sorted

townxelliot (approved):

service-front/composer.json
```diff
@@ -10,11 +10,12 @@
     ],
     "homepage": "https://github.com/ministryofjustice/opg-lpa",
     "require": {
+        "php": "^8.1",
```
townxelliot, line 13: Still a problem here, as you now have two entries for PHP.

nickdavis2001: Fixed now, anyway

townxelliot (approved):

PR #1482, opened 2023-04-05
LPAL-822 Update readme for better developer onboarding

townxelliot:
## Purpose

LPAL-822

## Approach

See ticket

## Learning

n/a

## Checklist


Makefile
```diff
@@ -64,11 +64,6 @@ dc-up: run-composers
 		--username AWS --password-stdin 311462405659.dkr.ecr.eu-west-1.amazonaws.com; \
 	docker-compose up
 
-# target for users outside MoJ to run the stack without 3rd party integrations
-.PHONY: dc-up-out
-dc-up-out: run-composers
-	@${MAKE} dc-up ORDNANCESURVEY=-
-
```
nickdavis2001, line 71: Is this no longer relevant then?

Makefile
```diff
@@ -164,24 +159,14 @@ dc-front-unit-tests:
...
-	docker build -f ./service-api/docker/app/Dockerfile -t integration-api-tests .;\
-	docker run -it --network="host" --rm integration-api-tests  sh -c "cd /app/tests/integration && php ../../vendor/bin/phpunit -v"
-
-.PHONY: test-pdf-local
-test-pdf-local:
-	docker build -f ./service-pdf/docker/app/Dockerfile -t pdf-tests .;\
-	docker run -d --env AWS_ACCESS_KEY_ID='-' --env AWS_SECRET_ACCESS_KEY='-' --name pdf-test-run pdf-tests; \
-	docker exec pdf-test-run /app/vendor/bin/phpunit -d memory_limit=256M; \
-	docker stop pdf-test-run;
-	docker rm pdf-test-run;
```
nickdavis2001, line 184: Is this defunct then?

README.md
```diff
@@ -25,32 +27,36 @@ cd opg-lpa
...
 * Whitespace and end of file fixers
 
 Add more as needed to the `.pre-commit-config.yaml`.
 
-For Linux users, revert to the instructions for installing phpcs, phpstan and pre-commit hooks:
-
-* <https://github.com/squizlabs/PHP_CodeSniffer>
-* <https://phpstan.org/user-guide/getting-started>
-* <https://pre-commit.com/index.html#install>
-
```
nickdavis2001, line 53: Is this just no longer relevant, or are we forgetting about running on linux for now?

README.md
```diff
@@ -199,28 +230,39 @@ The package.json in the root of the repo has all of the required dev dependancie
...
-docker run -v `pwd`/service-front/:/app/ composer update --prefer-dist --no-interaction --no-scripts --ignore-platform-reqs
+docker run -v `pwd`/service-front/:/app/ composer update --prefer-dist --no-interaction --no-scripts
 ```
 
 (replacing `service-front` with the path to the application component you are adding a package to; note that you'll need to do this for the following commands as well)
 
 Packages can be added with:
 
 ```bash
-docker run -v `pwd`/service-front/:/app/ composer require author/package --prefer-dist --no-interaction --no-scripts --ignore-platform-reqs
```
nickdavis2001, line 219: There is also in Makefile targets of run-front-composer  , run-api-composer etc.  At the mo' these ignore platform reqs. So maybe Makefile needs updating too?  Is it worth taking the docker command out of README and directing people to simply.   make run-front-composer for example  ,  in order to not have the docker command in 2 places which might drift out with each other?

Makefile
townxelliot: It's replaced by the pdf unit tests.

README.md
townxelliot: I'm taking out the Linux instructions. Anyone who is Linux savvy would be able to convert these instructions to do the installation.

README.md
townxelliot: This is specifically for adding a new package, which we don't do very often. I will remove any `--ignore-platform-reqs` flags.

Makefile
nickdavis2001: 👍

README.md
nickdavis2001: 👍

README.md
nickdavis2001: 👍

Makefile
townxelliot: No, because we have dependencies on more services (Notify, for instance), so you can't really do anything useful with the service if you aren't a member of MoJ.

README.md
townxelliot: The only problem with this is that there are no composer images which allow you to specify the PHP version. This means you could potentially be updating composer deps with the wrong PHP version.

townxelliot: Another issue is that removing `--ignore-platform-reqs` means that you have to have any required extensions installed in the image, too. The composer images don't have the intl extension, for example, which means that you can't update the package versions if `--ignore-platform-reqs` is removed.

Makefile
nickdavis2001: 👍

README.md
townxelliot: I am going to suggest this is left for a separate ticket. For now, I'll reinstate `--ignore-platform-reqs`.

townxelliot: I've figured it out. We use a multi-stage build, and locally use the first part of the build, leaving composer installed. When in CI, we let the whole Dockerfile run, which removes the composer files. It's pretty straightforward and non-intrusive, so I will have a go at implementing this approach this afternoon.

townxelliot: Raised LPAL-1196 with a suggestion to implement the dev image approach.

nickdavis2001: 👍

PR #1571, opened 2023-07-24
LPAL-1231 take apk upgrade out of api dockerfile.  

nickdavis2001:
same time

## Purpose

_Remove the apk upgrade from Dockerfile for api because this is leading to unwanted upgrades that cause performance problems.  Meanwhile change the aws keys for locally running dynamodb from hyphen which is not longer allowed, to devkey. Note - we ought to do a similar removal of apk upgrade in the service-front in future, but , on a seperate PR _

## Approach

_Update Dockerfile and docker-compose.yml_

## Checklist

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -111,8 +111,8 @@ services:
     volumes:
       - ./local-config/:/app/
     environment:
-      AWS_ACCESS_KEY_ID: "-"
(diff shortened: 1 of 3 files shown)


ntse (dismissed):

nickdavis2001:
Superseded

PR #1884, opened 2024-04-16
Update idea folder with working runners and configuration for PHPStorm

cooperaj:
## Purpose

Getting all the unit test runners and such working in PHPStorm

Still to be done for further fixes:

- Add Cypress test runners (custom python code used atm so that needs picking apart)
- Fixing up the debug enabling/disabling situation so we can run that locally a little easier without having to rebuild everything every time we want to turn it on.

## Checklist

* I have updated documentation (Confluence/GitHub wiki/tech debt doc) where relevant
* I have added mandatory tags to terraformed resources, where possible
* If I have a new OPG component dependency, I have updated the `metadata.json` with the repo location.
* If I added a package.json or composer.json, I also made sure this is included in the script in `.github/workflows/dependabot-update.yml`


Makefile
```diff
@@ -30,23 +30,23 @@ reset:
...
 
 .PHONY: run-pdf-composer
 run-pdf-composer:
-	@docker run -v `pwd`/service-pdf/:/app/ composer:${COMPOSER_VERSION} composer install --prefer-dist --no-interaction --no-scripts --ignore-platform-reqs
+	@docker run --rm -v `pwd`/service-pdf/:/app/ composer:${COMPOSER_VERSION} composer install --prefer-dist --no-interaction --no-scripts --ignore-platform-reqs
 
 .PHONY: run-api-composer
 run-api-composer:
-	@docker run -v `pwd`/service-api/:/app/ composer:${COMPOSER_VERSION} composer install --prefer-dist --no-interaction --no-scripts --ignore-platform-reqs
+	@docker run --rm -v `pwd`/service-api/:/app/ composer:${COMPOSER_VERSION} composer install --prefer-dist --no-interaction --no-scripts --ignore-platform-reqs
```
nickdavis2001, line 41: I think we had a --rm in there because otherwise containers get left around 🤔

service-api/docker/app/Dockerfile
```diff
@@ -46,11 +46,11 @@ COPY --chown=root:root shared/docker/app/php-fpm-logging.conf /usr/local/etc/php
 COPY --chown=root:root shared/docker/app/www.conf /usr/local/etc/php-fpm.d/
 
 # Migration script
-COPY --chown=root:root service-api/docker/app/db-migrations.sh /app/db-migrations.sh
+COPY --chown=root:root service-api/docker/app/db-migrations.sh /usr/local/bin/db-migrations.sh
 COPY --chown=root:root service-api/phinx.php /app/phinx.php
 COPY --chown=root:root scripts/containers/health-check.sh /usr/local/bin/health-check.sh
 
-RUN chmod +x /app/db-migrations.sh /usr/local/bin/health-check.sh
+RUN chmod +x /usr/local/bin/db-migrations.sh /usr/local/bin/health-check.sh
```
nickdavis2001, line 53: Why did this need to change?  No problem, just wondered

Makefile
cooperaj: This is me adding --rm. As those lines don't have one and my containers list was full of composer containers.

service-api/docker/app/Dockerfile
cooperaj: Because the docker-compose file now mounts the entire app folder in the db-migrations.sh file was getting "overwritten" since it didn't exist at that path in the mounted folder.

This just moves it out of the way so that doesn't happen.

Makefile
nickdavis2001: Whoops sorry way round , you're right :)

service-api/docker/app/Dockerfile
nickdavis2001: 👍

nickdavis2001 (approved):
lgtm :)

PR #2301, opened 2025-03-10
Remove deprecated library. Fix makefile.

mattmachell:
## Purpose

Remove deprecated library per [LPAL-1267](https://opgtransform.atlassian.net/browse/LPAL-1267)

[LPAL-1267]: https://opgtransform.atlassian.net/browse/LPAL-1267?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Makefile
```diff
@@ -34,15 +34,15 @@ run-front-composer:
...
+	@docker run --rm -v `pwd`/service-pdf/:/app/ composer:${COMPOSER_VERSION} composer install --prefer-dist --no-interaction --no-scripts --ignore-platform-reqs
 
 .PHONY: run-api-composer
 run-api-composer:
 	@docker run --rm -v `pwd`/service-api/:/app/ composer:${COMPOSER_VERSION} composer install --prefer-dist --no-interaction --no-scripts --ignore-platform-reqs
 
 .PHONY: run-admin-composer
 run-admin-composer:
-	@docker run --rm -v `pwd`/service-admin/:/app/ composer:${COMPOSER_VERSION} composer update tecnickcom/tcpdf --prefer-dist --no-interaction --no-scripts --ignore-platform-reqs
+	@docker run --rm -v `pwd`/service-admin/:/app/ composer:${COMPOSER_VERSION} composer composer install --prefer-dist --no-interaction --no-scripts --ignore-platform-reqs
```
nickdavis2001, line 45: This is good, fixes an incorrectly committed previous mod to Makefile

service-pdf/composer.json
```diff
@@ -14,11 +14,11 @@
         "laminas/laminas-filter": "^2.12.0",
         "laminas/laminas-barcode": "^2.11",
         "tecnickcom/tcpdf": "6.7.5",
-        "setasign/fpdi-tcpdf": "^2.0",
         "laminas/laminas-zendframework-bridge": "^1.1.1",
         "laminas/laminas-log": "^2.13",
         "symfony/validator": "^4.2",
-        "egulias/email-validator": "~1.2"
+        "egulias/email-validator": "~1.2",
+        "setasign/fpdi": "2.3.7"
```
nickdavis2001, line 21: A problem to be aware of is that this could all appear to work, all tests pass, UAT works, but resulting printed LPA won't scan. So this needs to be tested by scanning by hand LPA(s) generated from this UAT environment. This is why we don't currently have these packages upgraded. It looks like it ought to be a simple fix... but....

mattmachell: That's why I pinned it to the version provided by the old repo. The version of `setasign/fpdi` in the lock file is identical to the one we currently use.

nickdavis2001 (dismissed):

nickdavis2001 (approved):

PR #2378, opened 2025-05-20
LPAL-1273 replace laminas-log with monolog

nickdavis2001:
## Purpose

_Migrate from laminas-log which is abandonware, to Monolog as a replacement_

## Approach

_Monolog is somewhat of a drop-in replacement from laminas-log, however it needed to be injected throughout the system which led to a lot of changed files and tests, and changes to factories and module config_

## Checklist


service-admin/src/App/src/Logging/LoggerTrait.php
```diff
@@ -2,28 +2,24 @@
...
-    /**
-     * @return LaminasLogger $logger
-     */
-    public function getLogger()
+    public function getLogger(): LoggerInterface
     {
-        if (!$this->logger instanceof LaminasLogger) {
-            $this->logger = new LaminasLogger();
-            $this->logger->addWriter(new StreamWriter('php://stderr'));
+        if ($this->logger === null) {
```
cooperaj, line 20: Now it's all working is this bit neccessary?

service-api/module/Application/src/ControllerFactory/AuthControllerAbstractFactory.php
```diff
@@ -62,6 +63,12 @@ public function __invoke(ContainerInterface $container, $requestedName, array $o
 
         $controller = new $requestedName($authenticationService, $service);
 
+        $traitsUsed = class_uses($controller);
```
cooperaj, line 66: Might be worth adding a comment about how 'class_uses' does *not* return traits in parent classes. Works here but we did have an issue with that.

service-front/module/Application/src/Module.php
```diff
@@ -144,13 +142,27 @@ public function getServiceConfig()
...
                 'ApiClient'             => 'Application\Model\Service\ApiClient\ClientFactory',
                 'AuthenticationService' => 'Application\Model\Service\Authentication\AuthenticationServiceFactory',
                 'OrdnanceSurvey'        => 'Application\Model\Service\AddressLookup\OrdnanceSurveyFactory',
                 'SessionManager'        => 'Application\Model\Service\Session\SessionFactory',
                 'MailTransport'         => 'Application\Model\Service\Mail\Transport\MailTransportFactory',
+                'Logger'                => 'MakeShared\Logging\LoggerFactory',
+                'ExporterFactory'       => ReflectionBasedAbstractFactory::class,
+
+                'CsrfValidator' => function (ServiceLocatorInterface $sm) {
+                    $csrfName = 'secret_' . md5(get_class($this));
```
cooperaj, line 157: Did we figure out what "$this" resolved to in this context?

shared/module/MakeShared/src/Logging/LoggerTrait.php
cooperaj, line 1: I think I've seen this trait in the other projects? Does that mean it's been duplicated?

service-pdf/src/Opg/Lpa/Pdf/Traits/LoggerTrait.php
```diff
@@ -0,0 +1,28 @@
+<?php
+
+namespace Opg\Lpa\Pdf\Traits;
+
+use Monolog\Handler\StreamHandler;
+use Monolog\Level;
+use Monolog\Logger;
+use Psr\Log\LoggerInterface;
+
+trait LoggerTrait
```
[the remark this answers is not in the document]
nickdavis2001: Because pdf is just a console app that doesn't do injection.

shared/module/MakeShared/src/Logging/LoggerTrait.php
nickdavis2001: This is the version of the trait that works with injection, used by service-front and service-api.  The admin side and pdf don't do injection, so they have their own copies

service-admin/src/App/src/Logging/LoggerTrait.php
cooperaj: That works, but we should probably resolve that when we rework with Mezzio

cooperaj (approved):
Nice

PR #2567, opened 2025-09-15
Renovate ministryofjustice opg GitHub actions 4.x

andrewpearce-digital:


.github/workflows/workflow_path_to_live.yml
```diff
@@ -61,11 +61,11 @@ jobs:
         with:
           terraform_directory: "./terraform/region"
       - name: Bump version and push tag
-        uses: ministryofjustice/opg-github-actions/actions/semver-tag@9c80e2dfa5701930bcd444f26081c1183d951409 # v4.3.2
+        uses: ministryofjustice/opg-github-actions/actions/semver@9c80e2dfa5701930bcd444f26081c1183d951409 # v4.3.2
```
charlesmmarshall, line 64: not sure if its used, but the output var has changed to `tag` as well

charlesmmarshall: yeah, this line - https://github.com/ministryofjustice/opg-lpa/blob/main/.github/workflows/workflow_path_to_live.yml#L33 will need a change too

charlesmmarshall (approved):

PR #2586, opened 2025-09-19
Manual cypress install step.

mattmachell:


.github/workflows/cypress_tests.yml
```diff
@@ -102,6 +102,7 @@ jobs:
         run: |
           pip install boto3
           npm ci --ignore-scripts .
+          cypress install
```
gregtyler, line 105: I think you need either `npx cypress install` or `./node_modules/.bin/cypress install`

gregtyler (approved):

PR #2589, opened 2025-09-19
Test remove opcache preload

mattmachell:
Test to see impact of removing opcache preload.


mattmachell:
It causes a good few minutes wait on Make local startup before the app is usable, which makes change hard. It is likely also impacting container startup on deploy. Added this branch to test the impact of removal. May need some tweaking, but I thought I'd start with measuring full removal.

PR #2596, opened 2025-09-22
Make file api composer changes

mattmachell:
Add shortcuts. Test by updating symfony/validator


ramchale (approved):
We could reduce the repetition by setting the volume mount based on the target, but good to get these in and play with optimising later if we want :)

PR #3260, opened 2026-04-10
Add annotations from PHPUnit failures

gregtyler:
## Purpose

Fixes LPAL-2012 #patch

## Approach

Output a teamcity log—because it's the only one that contains line numbers—and use it with a problem matcher to detect where errors occurred and convert them into annotations.


gregtyler:
Closing this as the output isn't consistent enough from PHPUnit to extract the correct location from

PR #3283, opened 2026-04-16
LPAL-2019 bug fix - Preferences and instructions missing text when printing PDF

allenannom:
LPAL-2019 

the HTMLPurifier normalises \r\n to \n, and the PDF formatter splits text on \r\n to calculate the line breaks and the continuation sheet overflow.
When the text was stored with \n only, the formatter treated the entire content as a single line, causing the text to be cut off

now restoring \r\n after htmlpurifier in the filter, and make the formatter handle any new line style (\r\n, \r, \n)


nickdavis2001 (approved):

acsauk (approved):
:shipit: 

(with beefed up PDF diff checks to follow in future PR)

PR #3289, opened 2026-04-16
LPAL-1681: POC mezzio-csrf implementation

acsauk:
## Purpose

There are a couple of additional things in place due to running a mezzio component inside laminas-mvc, but this seems to work OK, and if we're happy with it, we could update all our handlers to use this.

The question remains though, whether we will see the same intermittent issues with inconsistent validation errors that appeared to be linked to higher traffic validations writing to a session - it's difficult to simulate this when applying to a single handler.

Satisfies LPAL-1681.

### DIFF
--- a/service-front/module/Application/src/Handler/TypeHandler.php
+++ b/service-front/module/Application/src/Handler/TypeHandler.php
@@ -17,6 +17,7 @@ use MakeShared\DataModel\Lpa\Document\Document;
 use MakeShared\DataModel\Lpa\Document\Donor;
 use MakeShared\DataModel\Lpa\Lpa;
 use Application\Helper\MvcUrlHelper;
+use Application\Middleware\CsrfValidationMiddleware;
 use Mezzio\Template\TemplateRendererInterface;
 use Psr\Http\Message\ResponseInterface;
 use Psr\Http\Message\ServerRequestInterface;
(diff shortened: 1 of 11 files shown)


service-front/module/Application/src/Handler/TypeHandler.php
```diff
@@ -49,9 +51,21 @@ public function handle(ServerRequestInterface $request): ResponseInterface
...
         $form = $this->formElementManager->get('Application\Form\Lpa\TypeForm');
 
+        // mezzio-csrf handles CSRF via CsrfValidationMiddleware; remove the form's
+        // built-in CSRF element so it doesn't interfere with form validation.
+        $csrfElement = $form->getCsrf();
+        if ($csrfElement !== null) {
+            $csrfName = $csrfElement->getName();
+            $form->remove($csrfName);
+            $form->getInputFilter()->remove($csrfName);
+        }
```
acsauk, line 65: This wouldn't be required when fully in mezzio. It's a workaround as our form implementation expects the CSRF field to follow the Laminas implementation, so we remove it when adding the mezzio implementation.

service-front/module/Application/src/Middleware/CsrfValidationMiddleware.php
```diff
@@ -0,0 +1,32 @@
...
+
+use Fig\Http\Message\RequestMethodInterface;
+use Laminas\Diactoros\Response\RedirectResponse;
+use Mezzio\Csrf\CsrfMiddleware;
+use Psr\Http\Message\ResponseInterface;
+use Psr\Http\Message\ServerRequestInterface;
+use Psr\Http\Server\MiddlewareInterface;
+use Psr\Http\Server\RequestHandlerInterface;
+
+class CsrfValidationMiddleware implements MiddlewareInterface
```
acsauk, line 15: This would remain when fully in mezzio

service-front/module/Application/src/Middleware/MvcSessionBridgeMiddleware.php
```diff
@@ -0,0 +1,46 @@
...
+ * This replaces SessionMiddleware (and PhpSessionPersistence) for routes that are
+ * still dispatched through the laminas-mvc stack, where the PHP session has already
+ * been started before our PSR-15 pipeline runs. Using the standard SessionMiddleware
+ * in this context causes a PHP warning because PhpSessionPersistence tries to change
+ * the session ID on an already-active session.
+ *
+ * On the way in:  wraps $_SESSION in a Mezzio Session and injects it as a request attribute.
+ * On the way out: writes any changes back to $_SESSION.
+ */
+class MvcSessionBridgeMiddleware implements MiddlewareInterface
```
acsauk, line 27: We would drop this when fully in mezzio.

acsauk:
Now implemented in #3395

PR #3291, opened 2026-04-16
LPAL-1966: Remove laminas-test

acsauk:
## Purpose

Fixes LPAL-1966


service-front/module/Application/tests/Model/Service/Lpa/ContinuationSheetsTest.php
```diff
@@ -4,18 +4,17 @@
...
 use Application\Model\Service\Lpa\ContinuationSheets;
 
-final class ContinuationSheetsTest extends AbstractHttpControllerTestCase
+final class ContinuationSheetsTest extends TestCase
 {
     private ContinuationSheets $service;
 
     public function setUp(): void
     {
-        parent::setUp();
```
acsauk, line 18: This is OK, as we aren't requiring the full laminas-mvc setup in any of the tests (config, service manager etc) which is what the parent class `setUp()` provided.

nickdavis2001 (approved):

gregtyler (approved):

PR #3296, opened 2026-04-17
Bump npm vulns

gregtyler:
## Purpose

Upgraded a bunch of transient dependencies that have vulnerabilities in them.

## Approach

Run `npm audit` to target some auto-patchable vulnerable packages (basically what dependabot does, but all at once)

#patch


acsauk (approved):
:shipit:

PR #3298, opened 2026-04-17
LPAL-1241: Add aria-labelledby and aria-modal to base dialog

acsauk:
## Purpose

Fixes LPAL-1241

### DIFF
--- a/service-front/module/Application/view/application/authenticated/dashboard/confirm-delete.twig
+++ b/service-front/module/Application/view/application/authenticated/dashboard/confirm-delete.twig
@@ -6,7 +6,7 @@
 
 <header class="popup-header">
     <i class="icon icon-bin"></i>
-    <h2 class="heading-small popup-header-title">Are you sure you want to delete this?</h2>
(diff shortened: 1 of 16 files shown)


service-front/module/Application/view/application/authenticated/dashboard/confirm-delete.twig
```diff
@@ -6,7 +6,7 @@
 
 <header class="popup-header">
     <i class="icon icon-bin"></i>
-    <h2 class="heading-small popup-header-title">Are you sure you want to delete this?</h2>
+    <h2 class="heading-small popup-header-title" id="dialog-title">Are you sure you want to delete this?</h2>
```
mattmachell, line 9: Do we ever have more than one popup at once? Just wondering if this might need to be digalog-title-{id} or not.

acsauk: ~~I've checked through the codebase and we don't do this by design anywhere.~~

This could potentially happen if session timeout appears while a user is filling in details in a popup. Luckily we use a unique template for that so I'll update with a proper id there.

mattmachell (approved):

PR #3300, opened 2026-04-17
Update update-find-cypress-specs to 1.54.12

gregtyler:
This updates a bunch of underlying dependencies, some of which are very outdated. This includes a vulnerable version of undici.

#patch


acsauk (approved):
:shipit:

PR #3301, opened 2026-04-17
LPAL-2022: Remove controllers and some listeners

acsauk:
## Purpose

I initially tried to remove more of the listeners but we rely on getting user details, identity and current route name for the ViewVariablesListener so I settled for finishing off separating listener and middleware logic to make the infal deletes easier.

Fixes LPAL-2022


allenannom (approved):
🚀

PR #3306, opened 2026-04-20
LPAL-1242: Prefer dialog element, included browser behaviour to custom JS

acsauk:
## Purpose

Fixes LPAL-1242, LPAL-1244

## Approach

Moved from a div-based to dialog-based popup implementation for the app. As the API for `<dialog>` includes `showModal()` and browsers automatically handle focus management, we could drop a lot of custom JS.


cypress/e2e/Session.feature
```diff
@@ -26,17 +26,11 @@ Feature: Session
...
-        Then my focus is within "session-timeout-form"
-        When I press shift+tab
-        Then my focus is within "session-timeout-form"
-        When I press shift+tab
-        Then my focus is within "session-timeout-form"
+        # Verify the popup is a native modal dialog - the browser enforces the
+        # focus trap, so we test our code: correct element type, opened as a modal,
+        # and that focus is placed on the heading on open (custom code in showWarning())
+        And "timeout-popup" is a modal dialog
+        And "timeout-dialog-title" is the active element
```
acsauk, line 33: Focus management is natively supported by `<dialog>`, so no need to test code we didn't write.

service-front/assets/js/opg/session-timeout-dialog.js
```diff
@@ -137,44 +131,10 @@ window.SessionTimeoutDialog = function (options) {
     this.hideWarning();
 
     // Keep the session alive
-    $.get(this.keepSessionAliveUrl).complete(function () {
+    $.get(this.keepSessionAliveUrl).always(function () {
```
acsauk, line 134: `complete()` was removed in jQuery 3.0; `always()` replaced it.

service-front/module/Application/view/layout/partials/timeout.twig
acsauk, line 1: If anyone is looking to test this and doesnt want to get stuck in the weeds with the many timeout values that control this, you can call `window.timeoutWarningDialog.showWarning()` in the browser console to force it showing.

nickdavis2001 (approved):

PR #3326, opened 2026-04-23
LPAL-1968 GOV UK Frontend Changes - updated classes and added md 

allenannom:
LPAL-1968 
## Purpose

spike to understanding effort involved in migrating fully to GOV UK Frontend - then updated classes which have an equivalent in the GOV UK Frontend


docs/govuk-frontend-migration-plan.md
```diff
@@ -0,0 +1,129 @@
...
+
+- Swap legacy classes for `govuk-*` equivalents.
+- Replace `show-hide-content` markup with `govuk-radios__conditional` + `data-aria-controls`.
+- Run the page in the browser; run the matching Cypress flow.
+
+### Phase 2 — Cypress + bespoke SCSS catch-up
+
+Once the templates above are migrated:
+
+- Update the two Cypress selectors: `div.form-group-error` → `.govuk-form-group--error`; `.error-summary-heading` → `.govuk-error-summary__title`.
```
mattmachell: We should probably move to a better identifier pattern in tests that's decoupled from visual styles, rather than just classes related to forms as part of this. Would avoid similar issues in future. i.e. `[data-form-errors] `

nickdavis2001 (approved):

PR #3328, opened 2026-04-23
Lpal 2007 try mezzio now with mezzio

acsauk:
## Purpose

**Note I branched from #3327 as it wasn't merged by the end of the day and requires changes introduced there.**

This is a fresh (untested) mezzio app with:
* Container DI wired up by referencing a `ConfigProvider` in the existing laminas-mvc app
* Middleware that runs on every request piped in

Next steps for this will be to hook up one existing handler (`HomeHandler` is a good candidate) and wire up to docker using patterns established for `mezzio-test-app`. I considered re-using the `mezzio-test-app` but it wasn't clear how much we had deviated from a "standard"  setup and I ran into a couple of autoload issues when trying to share Handlers outside the directory. If we can overcome the autoload issues then it may be cleaner to have the app outside `service-front`.

### DIFF
--- a/.idea/php.xml
+++ b/.idea/php.xml
@@ -1,5 +1,12 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <project version="4">
+  <component name="LaravelPint">
+    <laravel_pint_settings>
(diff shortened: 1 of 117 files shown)


acsauk:
Now implemented in #3390

PR #3329, opened 2026-04-23
LPAL-1877 - Add runbook for using cloudshell

andrewpearce-digital:
## Purpose

Document the process for using cloudshell instead of cloud9

Fixes LPAL-1877

## Approach

- add cloudshell run book
- remove cloud9 run book
- allow cloudshell VPC environment to use VPC endpoints

## Learning

- https://docs.aws.amazon.com/cloudshell/latest/userguide/welcome.html
- https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/security_group
- https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/security_group_rule
- https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_DescribeSecurityGroups.html


service-pdf/docker/app/.snyk
```diff
@@ -6,5 +6,5 @@ ignore:
   snyk:lic:composer:tecnickcom:tcpdf:((GPL-3.0_OR_LGPL-3.0)_OR_GPL-2.0_OR_GPL-3.0):
     - "*":
         reason: This will not be resolved
-        expires: 2026-04-23T14:57:38.314Z
+        expires: 9999-04-23T14:57:38.314Z
```
gregtyler, line 9: Nit: maybe worth changing the description to something like "this is an acceptable license"?

gregtyler (dismissed):

service-pdf/docker/app/.snyk
andrewpearce-digital: good suggestion. I've done that

gregtyler (approved):

PR #3342, opened 2026-05-01
LPAL-1936 added activity logging to admin

allenannom:
LPAL-1936

### DIFF
--- a/service-admin/src/App/src/Handler/AbstractHandler.php
+++ b/service-admin/src/App/src/Handler/AbstractHandler.php
@@ -101,7 +101,7 @@ abstract class AbstractHandler implements
     }
 
     /**
-     * @param ServerRequestInterface $request
+     * @param string|null $adminId The authenticated admin's user id
      * @param string $event Dotted event name (e.g. "admin.user.search")
      * @param string $message Human-readable message
      * @param array<string, mixed> $context Additional context fields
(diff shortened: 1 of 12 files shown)


service-admin/src/App/src/Handler/AbstractHandler.php
```diff
@@ -92,4 +99,31 @@ protected function getFlashMessages(ServerRequestInterface $request): array
...
+        array $context = [],
+        string $level = 'info'
+    ): void {
+        $adminUser = $request->getAttribute('user');
+
+        $baseContext = ['event' => $event];
+
+        if ($adminUser !== null) {
+            $baseContext['admin_id'] = $adminUser->id ?? null;
+            $baseContext['admin_email'] = $adminUser->email->address ?? null;
```
mattmachell: I'd avoid logging email if you already log ID, due to emails having PII by their nature i.e. firstname.lastname@org

service-admin/src/App/src/Handler/FeedbackHandler.php
```diff
@@ -75,6 +75,17 @@ public function handle(ServerRequestInterface $request): ResponseInterface
                     $queryParams = $request->getQueryParams();
                     $isExport = (array_key_exists('export', $queryParams) && $queryParams['export'] === 'true');
 
+                    $this->auditLog(
+                        $request,
```
mattmachell: You probably want to avoid logging the whole request here as it'll be the entire payload contents by my reading? If that is a search it might have PII in that'll end up in the logs. 

Edit: Oh, realised after reading the code further up this is passed to the AuditLogger to extract the user id. Might be better just to pass in the user id for clarty of the interface.

service-admin/src/App/src/Handler/SignOutHandler.php
```diff
@@ -18,6 +18,12 @@ class SignOutHandler extends AbstractHandler
      */
     public function handle(ServerRequestInterface $request): ResponseInterface
     {
+        $this->auditLog(
+            $request,
```
mattmachell: This should probably be the admin user ID not the request.

service-admin/src/App/src/Handler/SignInHandler.php
```diff
@@ -107,15 +107,33 @@ public function handle(ServerRequestInterface $request): ResponseInterface
                                     ])
                             );
                         } else {
+                            $this->getLogger()->info('Admin signed in', [
+                                'event' => 'admin.auth.sign_in',
+                                'admin_id' => $identity->getUserId(),
+                                'admin_email' => $email,
```
mattmachell: See above note re logging email.

nickdavis2001 (approved):

PR #3355, opened 2026-05-07
LPAL-1720 - Encrypt secrets with key that has a policy

andrewpearce-digital:
## Purpose

Control who can encrypt and decrypt secrets in aws secrets manager

Fixes LPAL-1720

## Approach

- use new key for secret encryption in region

## Learning

- https://github.com/ministryofjustice/opg-terraform-aws-kms-key


acsauk (approved):
:shipit:

PR #3357, opened 2026-05-07
LPAL-2065 added GDS Classes to custom Print.scss

allenannom:
LPAL-2065 
The custom made print.scss stylesheet is used when needing to print. simply added GDS classes as well so we keep the custom styling


acsauk (approved):
Is the idea that we come back and remove the other classes once we've dropped the old GDS styles?

allenannom:
> Is the idea that we come back and remove the other classes once we've dropped the old GDS styles?

Yeah, once we are confident we can remove the legacy packages I think that would be a good time to remove the other classes @acsauk

PR #3384, opened 2026-05-14
LPAL-2061: Prefer govuk frontend classes for buttons and links

acsauk:
## Purpose

From what I could see, there was no replacement for the toolkit `@button()` so I would have had to use a hex colour, which felt a bit wrong so I couldn't complete the initial intent of the ticket to upgrade helpers. Instead, I've applied the govuk frontend classes we should be using which has allowed me to drop the majority of the custom CSS we were carrying for buttons and links (I had to do links due to their uses with button groups). 

Fixes LPAL-2061


allenannom (approved):
🚀 can't see a difference in the updated baseline images. Looks good to me

PR #3387, opened 2026-05-15
LPAL-2077: Set environment_pipeline_tasks_config.json on dispatch

acsauk:
## Purpose

Fixes LPAL-2077

### DIFF
--- a/.github/workflows/locust_tests.yml
+++ b/.github/workflows/locust_tests.yml
@@ -7,19 +7,19 @@ defaults:
 on:
   workflow_dispatch:
     inputs:
+      environment:
+        description: "Terraform workspace / environment name (e.g. 3359lpal1896addaut)"
+        required: true
+        type: string
       users:
         description: "Number of users"
         default: "5"
-        required: true
+        required: false
         type: string
       run_time:
         description: "Test duration (e.g. 2m, 5m)"
         default: "2m"
-        required: true
-        type: string
-      front_url:
-        description: "URL of the frontend to run tests against"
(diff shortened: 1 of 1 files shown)


.github/workflows/locust_tests.yml
```diff
@@ -85,12 +79,60 @@ jobs:
...
           pip install -r scripts/pipeline/ci_ingress/requirements.txt
 
+      - name: Generate Terraform artifact config (dispatch only)
+        if: github.event_name == 'workflow_dispatch'
+        run: |
+          ENVIRONMENT="${{ inputs.environment }}"
+          ACCOUNT_ID="050256574573"
+          FRONT_URL="https://${ENVIRONMENT}.development.front.lpa.opg.service.justice.gov.uk"
+
+          FRONT_SG=$(aws ec2 describe-security-groups \
```
acsauk, line 89: I *think* this should work as we have the required perms in [org-infra](https://github.com/ministryofjustice/opg-org-infra/blob/aac41afcce9fc7dff11c92b6ba68b235c165f35f/services/make-a-lasting-power-of-attorney/modules/oidc/oidc_ingress_management.tf#L27)

andrewpearce-digital: if it doesn't work you might have to supply the VPC id too. I know with boto it can only find SGs in the default VPC and we don't use the default VPC anymore on make. We might want to make this a script or run a targeted terraform apply to generate the config file

andrewpearce-digital (approved):

PR #3391, opened 2026-05-19
LPAL-2097| Region - adding custom policies for application log group encryption key

radhiashah:
## Purpose

- Updating application_log_group kms_key with new custom policies attribute from kms module update. 
- Added permissions for cloudwatch to use kms key for encyrption on logs. 

Fixes LPAL-2097

## Approach

- Added custom kms key policies document with additional policy.
- Called new custom policy into key module via "additional_policy_documents" variable.

## Learning

`Multiple documents can be combined using the source_policy_documents or override_policy_documents attributes. source_policy_documents requires that all documents have unique Sids, while override_policy_documents will iteratively override matching Sids.
`
https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/iam_policy_document#example-with-both-source-and-override-documents

### DIFF
--- a/terraform/region/kms_policies_custom.tf
+++ b/terraform/region/kms_policies_custom.tf
@@ -1,20 +1,6 @@
 
-# custom application log group encryption key policy document
-
+# additional custom policy document for application_log_group kms_key
 data "aws_iam_policy_document" "application_log_group_kms_policy" {
-  statement {
-    sid    = "Allow root to manage application log key policies"
-    effect = "Allow"
-
-    principals {
-      type        = "AWS"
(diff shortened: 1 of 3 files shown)


terraform/region/kms_policies_custom.tf
```diff
@@ -0,0 +1,42 @@
+
+# custom application log group encryption key policy document
+
+data "aws_iam_policy_document" "application_log_group_kms_policy" {
+  statement {
```
andrewpearce-digital: This statement is already on the key

charlesmmarshall (approved):

PR #3395, opened 2026-05-21
LPAL-2046: Implement mezzio-csrf

nickdavis2001:
This PR establishes most of the request lifecycle end-to-end: session, authentication, CSRF, user details, and LPA loading.

I went as far as wiring up `/dashboard` and `/lpa-type` to prove CSRF was working.

Also added some commands to the Makefile for the mezzio app. Main QOL improvement is ensuring dev mode gives us a useful error page when things break like:

Screenshot 2026-05-27 at 18 32 22

I've made a judgment call on a couple of things:
- When setting classes in `dependencies.global.php`, prefer inline closures to factories if we're just getting services from the container. Use factories when any logic is involved so we can test it.
- We duplicate rather than reference classes under the legacy laminas-mvc app namespace `Application` (I've not updated a few large files in the PR as it was already too big, will make a follow-up for duplicating those over e.g. `Application\Model\Service\Lpa\Application`)


service-front/mezzio/src/App/src/ConfigProvider.php
```diff
@@ -38,7 +38,7 @@ public function getTemplates(): array
     {
         return [
             'paths' => [
-                'app'         => [__DIR__ . '/../templates/app'],
+                //'app'         => [__DIR__ . '/../templates/app'],
```
nickdavis2001: had to comment this out for now for the site to work

allenannom (dismissed):

service-front/mezzio/src/App/src/ConfigProvider.php
gregtyler: There's no `app` folder, so I've hard-removed this instead

allenannom (dismissed):

allenannom (approved):
🚀

PR #3397, opened 2026-05-21
LPAL-2061: Add guidance regression test, refine CSS build

acsauk:
## Purpose

I noticed I broke guidance (again) when messing with classes on another branch so I'm getting this in before finalising that one to prove my changes are valid.

Enables LPAL-2061

## Approach

I was getting an error with the minifying step since a recent merge to main so decided to rely on sass for minifying as we already use it in the css build script.


service-front/assets/sass/print.scss
```diff
@@ -66,8 +66,7 @@
 
     .table-summary-heading {
         .govuk-heading-s,
```
acsauk, line 68: Esbuild minifying was hiding the fact we had a duplicate class here as it drops duplicates in minification.

service-front/public/assets/v2/css/print.min.css
```diff
@@ -1 +1 @@
...
...nlinks *,.summary-print main>hr *,.summary-print main>p *,.summary-print .govuk-button *,.summary-print .button *,.summary-print .button-link *,.summary-print .no-print-for-summary *,.summary-print a:after{display:none !important}.summary-print #content{padding:20px}.summary-print .table-summary{text-align:left;display:block;margin-top:40px;float:none}.summary-print .table-summary tbody,.summary-print .table-summary tr{display:block}.summary-print .table-summary th,.summary-print .table-summary td{display:block;width:100%}.summary-print .table-summary tr:first-of-type .table-summary-title{font-size:24px}.summary-print .table-summary-title{font-weight:700;border-bottom:0 !important;padding-bottom:0 !important;margin-bottom:0 !important}.summary-print .table-summary-heading .govuk-heading-s,.summary-print .table-summary-heading .heading-small{font-size:24px;margin-top:10px;margin-bottom:10px}.summary-print .group-single,.summary-print .group-element,.summary-print .group-end{float:none;page-break-after:auto;break-after:auto}.summary-print .group-single td,.summary-print .group-element td,.summary-print .group-end td{padding-bottom:10px;margin-bottom:10px;border-bottom:1px solid gray}
```
acsauk, line 1: Hard to parse but the changes here are spacing before `!important`.

allenannom (approved):
🚀

PR #3403, opened 2026-05-22
LPAL-2047: Add mezzio sessions

allenannom:
LPAL-2047
- duplicated handlers to mezzio app
- installed packages for mezzio sessions
- refactored handlers that references session utility to now use mezzio session
- added test for LoginHandler.php


acsauk (approved):
:shipit:

PR #3406, opened 2026-05-26
Bump idna

gregtyler:
## Purpose

Upgrade to non-vulnerable version

#patch

## Approach

```sh
uv pip compile requirements.in -o requirements.txt --upgrade-package idna
```

Three of these were easy, but I also updated locust to use `uv` like the other packages which also required changes to the in file to match the latest updates. This completely changed the dependency block, which I presume is from tidying up things that have slipped over the last few years.


andrewpearce-digital (approved):

acsauk (approved):
:shipit:

PR #3408, opened 2026-05-27
LPAL-1585 - Count suspicious URI patterns at WAF

andrewpearce-digital:
## Purpose

Prevent scanning noise and probing.

Fixes LPAL-1585

## Approach

- add regex for suspicious URI patterns
- set to count mode for evaluation

## Learning

- https://github.com/ministryofjustice/opg-use-an-lpa/blob/main/terraform/account/region/waf.tf


acsauk (approved):
:shipit:

PR #3409, opened 2026-05-28
Ensure tokens are always 20 characters long

gregtyler:
## Purpose

If the generated token starts with 0s, `gmp_strval` will generate a string shorter than 20 characters:

```php
echo gmp_strval('0xffff', 62) === 'H31';
```

The string is no less random (10% of numbers under 1,000,000 are 6 numbers of fewer) but might cause confusion and occasionally causes tests to fail.

Fixes LPAL-1520 #patch

## Approach

Therefore, if the token string is less than 20 characters we pad it with 0s to make them all 20 chars long.


allenannom (approved):

acsauk (approved):
:shipit:

PR #3412, opened 2026-05-29
LPAL-2048 wired up unathenticated routes

allenannom:
LPAL-2048


service-front/mezzio/src/App/src/Service/Date/DateService.php
```diff
@@ -0,0 +1,20 @@
+<?php
+
+declare(strict_types=1);
+
+namespace App\Service\Date;
+
+use DateTime;
+
+class DateService
```
gregtyler, line 9: Could we pull in the existing services rather than duplicating them? It seems confusing to have two copies, and I worry about how easily we could drift. For example, if the Laminas DateService gets a fix added to it and we forget to backport it here then we'll end up getting the same bug twice.

Appreciate we can't do this with handlers because they're so tied to the framework, but the benefit of abstracting to services is that they should be reusable by both Laminas and mezzio.

service-front/mezzio/src/App/src/Service/Feedback/FeedbackService.php
```diff
@@ -0,0 +1,40 @@
...
+                'status' => $ex->getStatusCode(),
+                'exception' => $ex,
+            ]);
+
+            if ($ex->getStatusCode() === 400) {
+                throw new FeedbackValidationException($ex->getMessage());
+            }
+
+            throw $ex;
+        }
```
gregtyler, line 46: This seems to be missing a bunch of functionality around sending emails on submission?

service-front/mezzio/src/App/src/Service/System/StatusService.php
```diff
@@ -0,0 +1,59 @@
...
+    public function check(): array
+    {
+        $api = $this->checkApi();
+
+        $status = $api['ok'] ? Constants::STATUS_PASS : Constants::STATUS_FAIL;
+
+        return [
+            'ok' => $api['ok'],
+            'status' => $status,
+            'api' => $api,
```
gregtyler: This class has lost support for other services (DynamoDB, Redis, etc)

service-front/mezzio/src/App/src/Service/Date/DateService.php
allenannom: It is quite messy for sure as we have duplicate code but since the plan is for the mezzio app to fully replace the existing one, we'd need to duplicate these services eventually anyway — doing it now means when we switch over, everything's already self-contained and we can just delete the legacy code without untangling shared references.
For the interim where both exist — any hotfixes would need applying to both, but the test suites on each side should catch if one drifts. It's a bit messy short-term, but it keeps the mezzio app cleanly decoupled and makes the final cutover simpler.

nickdavis2001 (dismissed):

nickdavis2001 (approved):

PR #3415, opened 2026-06-01
LPAL-1677 remove obsolete polyfill

nickdavis2001:
## Purpose

_Removal of old code for ease of maintenance. moj.polyfill.js is a details tag polyfill for IE/old Firefox. All supported browsers natively implement details tag. The code checks typeof document.createElement(tag).open === 'boolean' this will always be true nowadays, so the non-native polyfill branch never runs. The ARIA attribute injection still runs, but govuk-frontend already handles this for elements with data-module="govuk-details". Some older templates use details tag without that attribute though, so there's a partial overlap. The only remaining work it does (ARIA injection on non-govuk-frontend details tag) could be considered obsolete.._

## Approach

_Remove moj.polyfill.js_


nickdavis2001:
Looks like it is not possible at this point to disentangle this. Perhaps will be more straghtforward later once the rest of the frontend work is done

PR #3421, opened 2026-06-03
LPAL-1585 - reducing general scanning noise part 3

andrewpearce-digital:
## Purpose

Prevent behavior that tries to identify and exploit known vulnerabilities

Fixes LPAL-1585

## Approach

- Reorder suspicious URI rules
- Add a count rule for further suspiscous URIs
- add general high threshold rate limit rule in count mode


.github/workflows/phpunit.yml
```diff
@@ -54,3 +54,5 @@ jobs:
           name: ${{ matrix.scan.service }}-coverage-html
           path: ./coverage-html/
           overwrite: true
+
+# testing PHP run locally - radhia
```
radhiashah: remember to remove this - ( i added it so could push your branch to test php stuff)

andrewpearce-digital: done

radhiashah (approved):

PR #3424, opened 2026-06-03
LPAL-2066 and LPAL-2068 Drop old toolkit dependencies and fix remaining issues

nickdavis2001:
## Purpose

_Migration from legacy toolkit to govuk-frontend - final removal of dependency on old components and fixing of any resulting issues_

## Approach

CSS / Sass migration
--------------------------------------
- build-css.sh: removed --load-path flags for govuk_frontend_toolkit and
  govuk-elements-sass; removed copy of govuk_template_mustache print/base CSS
  (govuk-frontend.min.css is now the sole vendor CSS copied).
- package.json / package-lock.json: removed govuk_frontend_toolkit,
  govuk_template_mustache and govuk-elements-sass dependencies.
- assets/sass/settings/_variables.scss: removed legacy variable references.
- assets/sass/application.scss: minor cleanup.
- assets/sass/extensions/elements/_helpers.scss (new content): adds the em()
  Sass function and base element resets (heading margin/padding zero, etc.) that
  were previously supplied by govuk-elements-sass. Required so existing patterns
  that call em() continue to work.
- assets/sass/extensions/elements/_elements-typography.scss: updated for
  govuk-frontend variables.
- assets/sass/extensions/elements/_icons.scss: icon size/position fixes.
- assets/sass/patterns/_form-popup.scss: minor cleanup.
- assets/sass/patterns/_layout.scss (new): content-width / layout helpers that
  were previously provided by the toolkit.
- assets/sass/patterns/_lpa-list.scss: spacing fixes.
- assets/sass/patterns/_table-summary.scss: spacing fixes (uses em() helper).

Twig template fixes 
-----------------------------------------------
Minor class/markup corrections in several templates following the CSS
changes: correspondent, how-primary-attorneys-make-decision,
when-replacement-attorney-step-in, reuse-details, type, cookies, feedback, and
the shared layout.twig. 

Mezzio form fixes
------------------
- AbstractMainFlowForm.php, BlankMainFlowForm.php: still had old button class , so changed to govuk-button.

Makefile
--------
- Removed references to govuk_frontend_toolkit node_modules path in dc-build-clean
  and hard-reset-front targets.
- cypress-open: updated CYPRESS_baseUrl from port 7002 to 7004 (Mezzio local port).

Updated baselines
------------------
- Baseline files updated to reflect minor differences


service-front/assets/sass/extensions/elements/_helpers.scss
```diff
@@ -1,6 +1,88 @@
...
+main {
+    @include govuk-font(19);
+}
+
+// govuk-elements-sass/_govuk-template-base.scss set strong/b to weight 600.
+// Without it, browsers default to 700 (bolder) for <strong>/<b>.
+b,
+strong {
+    font-weight: 600;
+}
```
gregtyler: Is this needed? It feels weird to have slightly different font weights to other gov.uk websites (I also don't think it makes a difference since we only supply a 400 and a 700 font through woff files)

service-front/assets/sass/patterns/_details.scss
```diff
@@ -1,10 +1,81 @@
...
+// =======================================================
+// DETAILS / SUMMARY
+//
+// Restores the custom govuk-elements-sass arrow + link styling for plain
+// <details>/<summary> elements that don't use the .govuk-details component
+// class (e.g. checkout "What do I need to do next?", type page help, etc.).
+// govuk-frontend only styles .govuk-details__summary; without these rules the
+// browser shows its native disclosure triangle with no colour or underline.
+
+details:not(.govuk-details) {
```
gregtyler: Rather than creating custom CSS that we'll need to maintain, could we just update the remaining plain `<details>` elements to use `govuk-details`?

service-front/assets/sass/patterns/_layout.scss
```diff
@@ -1,6 +1,31 @@
 // =======================================================
 // LAYOUT
 
+// Width constraint for the main content element.
+// Previously applied via govuk-elements %site-width-container on #content.
+// Mirrors the govuk-frontend .govuk-width-container behaviour.
```
gregtyler: If this mirrors govuk-width-container, could we just add that class in the template rather than adding custom CSS?

service-front/assets/sass/extensions/elements/_helpers.scss
nickdavis2001: Probably historical . Have removed this with no problems

service-front/assets/sass/patterns/_details.scss
nickdavis2001: OK, thanks for the suggestion, done :)

service-front/assets/sass/patterns/_layout.scss
nickdavis2001: Ok, done :)

gregtyler (approved):

PR #3443, opened 2026-06-08
LPAL-2071 remove old unused MOJ frontend

nickdavis2001:
## Purpose

_As part of migration to govuk frontend, remove references to unused MOJ frontend_

## Approach

_Remove it from service-admin which is the only service it turned out to be in. It was pulled in but never used_


allenannom (approved):
🚀

PR #3451, opened 2026-06-10
LPAL-2049 wired authentication handlers and routes

allenannom:
LPAL-2049

Ported and wired auth routes
Ported forms and helper classes
Added phpunit tests for each class/handler ported


service-front/mezzio/config/autoload/dependencies.global.php
acsauk: I know this wasn't part of the PR, but I think we should stick with imports vs qualifiers for readability in this file.

service-front/mezzio/config/autoload/dependencies.global.php
```diff
@@ -638,18 +645,24 @@ public function healthcheck(): array
             ),
             StatusService::class => static function (ContainerInterface $c): StatusService {
```
acsauk: Now we have some logic here we should probably move this to a factory so we can test it (as we won't be writing tests for this file)

service-front/mezzio/src/App/src/Filter/StripTagsPreservingAngleBrackets.php
acsauk, line 1: We need a test added for this ported file (and the other `Form` classes below).

service-front/mezzio/src/App/src/Handler/Lpa/CertificateProvider/CertificateProviderAddHandler.php
```diff
@@ -96,20 +96,18 @@ public function handle(ServerRequestInterface $request): ResponseInterface
...
                     ]);
 
                     return new RedirectResponse($reuseDetailsUrl);
                 }
             }
         }
 
-        /** @var \Application\Form\Lpa\AbstractActorForm $form */
-        $form = $this->formElementManager->get('Application\Form\Lpa\CertificateProviderForm');
+        /** @var \\App\\Form\\Lpa\\AbstractActorForm $form */
```
acsauk: Some weird formatting on the namespace here (appears in the other Handlers below as well)

service-front/mezzio/composer.json
```diff
@@ -90,15 +92,13 @@
     "autoload": {
         "psr-4": {
             "App\\": "src/App/src/",
-            "Application\\": "../module/Application/src/",
             "Alphagov\\Pay\\": "../module/Alphagov/Pay/src/",
```
acsauk, line 94: Do you think its time we port over this to the mezzio app as well? Or does it fit with another ticket better?

allenannom: I think this can be ported over as well. But conscious about how big the PR is already, I think another PR for that would be best

acsauk: Yup yup make sense 👍

acsauk (dismissed):
:shipit:

acsauk (approved):
:shipit:

PR #3452, opened 2026-06-10
LPAL-2068 add govuk-body class to paragraphs across the system

nickdavis2001:
## Purpose

_migration to govuk-frontend_

## Approach

_add a class of govuk-body to all paragraphs that didn't already have it. baseline screenshots updated (minor differences of only 10 pixels). New baseline files with different name for those that have the new footer appearance, similar to Use an LPA_


nickdavis2001:
Superseded by https://github.com/ministryofjustice/opg-lpa/pull/3459 which adds further changes to these

PR #3462, opened 2026-06-12
LPAL-2164: Prefer govuk-!-margin-bottom-3 to form-group

acsauk:
## Purpose

Reinstating the margins that `form-group` gave us so there is sufficient spacing between buttons and links on add another pages.

Fixes LPAL-2164


nickdavis2001:
Is it worth at this point looking into seeing if we can immediately use a class from govuk-frontend instead of the old toolkit to fix this?

nickdavis2001 (commented):
🏆

nickdavis2001 (approved):

PR #3469, opened 2026-06-15
LPAL-2168 headings to govuk-frontend 

nickdavis2001:
## Purpose

_Migration to govuk-frontend_

## Approach

_Headings changed to use new govuk-frontend classes. Margins maintained to match how it was with the old classes. Baselines updated where needed. form popup scss modified to support both old toolkit and govuk frontend to support follow on work from this while we're in a state of flux_


cypress/regressions/baseline/add-replacement-attorney-form.png
acsauk, line 1: Looks like the title alignment div is justified to top rather than centre here

cypress/regressions/baseline/checkout.png
acsauk, line 1: Possibly the same here? Looks like the summary title is more justified to the top rather than closer to the hr

cypress/regressions/baseline/lpa-type.png
acsauk, line 1: Margin between inset text and coloured heading here has been dropped

nickdavis2001: Hopefully now fixed

cypress/regressions/baseline/checkout.png
nickdavis2001: Hopefully now fixed

cypress/regressions/baseline/add-replacement-attorney-form.png
nickdavis2001: Hopefully now fixed

acsauk (approved):
:shipit:

PR #3476, opened 2026-06-16
LPAL-2139: Cypress mezzio tests (AKA fix everything we missed)

acsauk:
## Purpose

Fixes LPAL-2139


service-front/mezzio/src/App/src/Handler/Lpa/HowPrimaryAttorneysMakeDecisionHandler.php
```diff
@@ -66,7 +66,11 @@ public function handle(ServerRequestInterface $request): ResponseInterface
                 $postData = [];
             }
 
-            if ($postData['how'] != PrimaryAttorneyDecisions::LPA_DECISION_HOW_DEPENDS) {
+            // Use null coalescing rather than direct array access: unlike the legacy app's
+            // Laminas\Stdlib\Parameters (which returns null for missing keys silently),
+            // PSR-7 getParsedBody() returns a plain PHP array where accessing an absent key
+            // throws an E_WARNING promoted to ErrorException by the error handler.
+            if (($postData['how'] ?? null) != PrimaryAttorneyDecisions::LPA_DECISION_HOW_DEPENDS) {
```
acsauk, line 73: This is behind quite a few of the fixes in handlers and forms where in the legacy laminas-mvc app we got null for an unchecked checkbox.

service-front/mezzio/src/App/src/Middleware/CsrfValidationMiddleware.php
```diff
@@ -28,18 +32,33 @@ public function process(ServerRequestInterface $request, RequestHandlerInterface
...
+        // This prevents XHR sub-requests from overwriting the token that was embedded in
+        // the page's main form during the initial full-page load.
+        if ($session instanceof SessionInterface && $session->has(self::CSRF_KEY)) {
+            $token = $session->get(self::CSRF_KEY);
+        } else {
+            $guard = $request->getAttribute(CsrfMiddleware::GUARD_ATTRIBUTE);
+            $token = $guard->generateToken();
+        }
+
+        $request = $request->withAttribute(self::TOKEN_ATTRIBUTE, $token);
```
acsauk, line 61: This is a change to our current implementation for CSRFs - I wasn't able to get Cypress tests to pass using a per-request/single use token so Claude suggested using session-based CSRF tokens.

From what I understand, this isn't meaningfully less secure as it implements the [Synchronizer Token Pattern](https://cheatsheetseries.owasp.org/cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.html#synchronizer-token-pattern) that suggests the token should be unique per session rather than request (our current implementation). The token is still generated in the same way and we still have Same-Origin Policy applied so I'm fairly comfortable with the change but happy for any pushback if people can see issues with the approach.

service-front/mezzio/src/App/src/View/Twig/LegacyCompatExtension.php
```diff
@@ -474,35 +493,83 @@ public function formRadio(ElementInterface $element): string
         $currentValue = $element->getValue();
         $html         = '';
 
-        // Base attributes from the element (excluding value/type which vary per option)
-        $baseAttrs = array_diff_key($element->getAttributes(), array_flip(['value', 'type', 'id']));
-        $baseAttrs['type']  = 'radio';
-        $baseAttrs['class'] = $baseAttrs['class'] ?? 'govuk-radios__input';
+        // Determine the wrapper div class from element-level div-attributes, or default.
```
acsauk, line 497: Gotta be honest, I just let Claude go wild with this one. Its hard to parse if this is the correct approach as all of these functions are to replace built in laminas functions from the legacy app.

service-front/docker/web/etc/nginx/templates/default.conf.template
```diff
@@ -148,6 +156,10 @@ server {
 
         fastcgi_param SCRIPT_FILENAME /app/public$fastcgi_script_name;
 
+        # Tell PHP whether the original connection was HTTPS, based on what the
+        # upstream SSL terminator (AWS ALB, CloudFront, or local SSL proxy) sets.
+        fastcgi_param HTTPS $fastcgi_https;
+
```
acsauk, line 162: Some background: I ran into a problem here when trying to support docker Cypress runs and accessing the app from a browser. The return URL is hard-coded in the support JS that the pay mock uses, which would mean we would need to choose either localhost or mock-pay as host names, limiting the feature to working either in tests or via the browser. Additionally, the mock uses http vs https so we needed a way to determine which scheme was used and what should be used when redirecting requests. See local-ssl/docker/nginx.conf.template for the other change.

I've thought about this from a practical and security perspective, and I'm confident these changes don't introduce risk because:

* AWS ALB rewrites X-Forwarded-Proto to reflect the actual client-to-ALB protocol. A client cannot spoof it 
* drop_invalid_header_fields is set to true on the AWS ALB, and even without that, AWS rewrites this specific header at the load balancer level
* Forwarding $host is the standard nginx pattern — it reflects the hostname the client actually connected to, which the SSL proxy validated through its TLS certificate handshake

cypress.config.js
```diff
@@ -163,13 +163,20 @@ module.exports = defineConfig({
   numberOfGuidanceHelpTopics: 22,
   defaultCommandTimeout: 12000,
   requestTimeout: 12000,
+  pageLoadTimeout: 30000,
+  chromeWebSecurity: false,
```
acsauk, line 167: Needed this to support http redirects in govuk pay

cypress/mezzio-regressions/baseline/add-applicant.png
acsauk, line 1: I've copied these from the legacy app baselines. I plan to do another PR with updates to baselines so I can easily compare them and fix any issues. We may have some subtle differences as we've updated the twig extension functions, and it will be hard to see otherwise.

allenannom (approved):
🚀

PR #3477, opened 2026-06-16
LPAL-2170 migrate form labels etc from old toolkit to govuk-frontend

nickdavis2001:
## Purpose

_migration to govuk-frontend_

## Approach

- All incidences of form-label changed to govuk-label. 
- baseline files updated as needed to support the above_
- All incidences of form-hint changed to govuk-hint. 
- All incidences of bold changed to govuk-frontend equivalent
- All incidences of visually-hidden replaced with govuk-visually-hidden


service-front/assets/sass/print.scss
```diff
@@ -12,7 +12,7 @@
     }
 
     .govuk-visually-hidden,
-    .visually-hidden,
+    .govuk-visually-hidden,
     .govuk-visually-hidden,
```
mattmachell, line 14: You can merge the three different `.govuk-visually-hidden,` selectors here as it's just duplication

nickdavis2001: thanks good catch, now fixed

allenannom (approved):
🚀

PR #3489, opened 2026-06-17
LPAL-2178 injecting the identity email when the caller is fetching their own profile

allenannom:
Bug fix: Admin user search by ID returned the admin's own account instead of the searched user.

UserController::get() was unconditionally overwriting the returned email with the authenticated caller's email from their JWT identity. When an admin searched by user ID, the looked-up user's email was replaced with the admin's email before being returned — causing the subsequent email lookup to find the admin's own account. Fixed by only injecting the identity email when the caller is fetching their own profile ($identity->id() === $id)


nickdavis2001 (approved):

acsauk (approved):
:shipit:

PR #3491, opened 2026-06-18
LPAL-2182 migrate various classes to govuk-frontend

nickdavis2001:
## Purpose

_migration to govuk-frontend_

## Approach

_migrate alert to govuk-notification-banner or govuk-inset-text where appropriate
_migrate panel / panel-border-narrow / panel-border-thin / panel-border-wide to govuk-inset-text (informational) or govuk-warning-text (warnings) 
_migrate block-label to govuk-label or govuk-radios__label or govuk-checkboxes__label
_migrate form-control to govuk-input
_update baseline files where needed


cypress/regressions/baseline/when-lpa-starts.png
allenannom, line 1: this baseline has over a 100 pixels difference - could be the margin of the Inset. I'd suggest removing the inner div and p tag when using the `govuk-inset-text` class similar to the example here: https://design-system.service.gov.uk/components/inset-text/

allenannom (approved):

PR #3500, opened 2026-06-19
LPAL-2002 Make lock files for python with uv

allenannom:
Replaces all pip/pip3 installs with uv lock files. Every uv.lock records a SHA256 hash for every artifact — uv sync --locked verifies those hashes automatically on install, protecting against supply-chain attacks.

- All 4 workflows (workflow_path_to_live, start_task, cypress_tests, locust_tests) — `actions/setup-python + pip install` replaced with astral-sh/setup-uv (SHA-pinned) +` uv sync --locked`
- Makefile python-api-venv target` python3 -m venv + pip install` replaced with `uv sync --locked`
- New workflow **analysis-uv-lock-check.yml** — `runs uv lock --check `on all 8 projects on every PR touching a .toml or uv.lock file to make sure there is no drift. protects the lock from going stale if the pyproject.toml was edited and bumped a version or adds dependency but forgets to run `uv lock` before pushing
- Deleted all requirements.in/txt and setup.py


.github/workflows/cypress_tests.yml
```diff
@@ -61,15 +61,18 @@ jobs:
           role-duration-seconds: 900
           role-session-name: OPGLPACypressTestsSecurityGroupStart
 
-      - name: Setup Python
-        uses: actions/setup-python@e797f83bcb11b83ae66e0230d6156d7c80228e7c # v6.0.0
+      - name: Install uv
+        uses: astral-sh/setup-uv@fac544c07dec837d0ccb6301d7b5580bf5edae39  # v8.2.0
```
andrewpearce-digital, line 65: we should specify the version of python in uv action for the `uv python install` command later https://github.com/astral-sh/setup-uv#python-version, or pin to the exact version in the uv lock files. I think that version should be something like 3.14.4

andrewpearce-digital (approved):

PR #3505, opened 2026-06-23
LPAL-2163: Enable mezzio cypress test suite for PR and merge queue

allenannom:
LPAL-2163 
Added mezzio cypress test suite to PR and merge queue


.github/workflows/workflow_merge_queue.yml
```diff
@@ -273,7 +273,20 @@ jobs:
...
+
+  cypress_tests_Mezzio:
+    name: Run Cypress tests - Mezzio
+    uses: ./.github/workflows/cypress_tests.yml
+    needs:
+      - preprod_terraform_outputs
+      - run_preproduction_seed_db_task
+    with:
+      admin_url: https://${{ needs.preprod_terraform_outputs.outputs.admin_fqdn }}
+      front_url: https://${{ needs.preprod_terraform_outputs.outputs.front_fqdn }}
```
acsauk: We'll need to point towards the mezzio app FQDN here, but I'm not sure its available with the current outputs? @radhiashah do we need to add to the outputs now you've added the docker image step?

.github/workflows/workflow_merge_queue.yml
```diff
@@ -273,7 +273,20 @@ jobs:
...
+      account_id: "987830934591"
+      cypress_tags: "@FrontMezzio or @MezzioHome"
```
acsauk: This will need to mirror the above tags, as we don't tag the mezzio tests any differently from the laminas-mvc app, its the URLs that differ.

.github/workflows/workflow_merge_queue.yml
andrewpearce-digital: when `"mezzio_frontend_enabled": true,` is set, the FQDN front_url will be the mezzio app

acsauk (approved):

PR #3526, opened 2026-06-24
LPAL-2190 remove old flush--top from accordion header elements

nickdavis2001:
## Purpose

_migration from old toolkit to govuk-frontend_

## Approach

_accordion-header itself is a custom class not from the toolkit.  where accordion-header has a flush--top class,  remove that as flush--top comes from the old toolkit . replace this functionality by adding margin -top to _accordion.scss. lpa-type page needed an extra govuk-!-margin-bottom-6 to look OK. 1 baseline needed to be updated


acsauk (approved):
:shipit:

PR #3535, opened 2026-06-25
LPAL-2189 migrate text etc to govuk frontend

nickdavis2001:
## Purpose

_migration to govuk-frontend_

## Approach

- remove all incidences of text
- where needed, replace this with govuk-grid-row and add < div > within it, with class="govuk-grid-column-two-thirds", to keep the wrapping (it would go full width otherwise)
- migrate heading-xlarge and heading-large
- 1 reference to old bold class that does nothing, removed from person-form.js
- update baselines- a LOT have changed due to subtle changes in word wrapping - note that the life sustaining treatment one is now how it should be - wrapping had gone awry in a previous PR
- fixed Add Applicant cypress test to explicitly click named button rather than "occurence 1" which stopped working after these changes


cypress/regressions/baseline/fee-reduction.png
allenannom, line 1: I think the 'these means-tested benefits' link might not have the govuk link styling

cypress/regressions/baseline/instructions-preferences.png
allenannom, line 1: could we increase the text field width to match the existing one?

cypress/regressions/baseline/lpa-summary.png
allenannom, line 1: difference here in the paragraph width - could we adjust it to be the same as the existing width?

cypress/mezzio-regressions/baseline/add-applicant.png
allenannom, line 1: Could we indent the Attorneys checkbox options to be similar to previous baseline image?

gregtyler: It looks like the new one is using [Inset text](https://design-system.service.gov.uk/components/inset-text/) rather than [Conditional radios](https://design-system.service.gov.uk/components/radios/#conditionally-revealing-a-related-question), but I can't immediately track down where that's changed

nickdavis2001: Done :)

cypress/regressions/baseline/lpa-summary.png
nickdavis2001: Now fixed (in the mezzio regressions screenshot, not this one as this no longer needed)

cypress/regressions/baseline/fee-reduction.png
nickdavis2001: Fixed in mezzio-regressions - there is no file in the PR, because now no difference from the old one

cypress/regressions/baseline/instructions-preferences.png
nickdavis2001: This is now fixed but in the mezzio-regressions version.

allenannom (approved):
🚀

PR #3543, opened 2026-06-26
Remove direct dependency on hamcrest/hamcrest-php

gregtyler:
## Purpose

Hamcrest provides some additional assertion methods, but these are all things we can already do with PHPUnit. Replace our few usages with the PHPUnit equivalents and remove the direct dependency on hamcrest in shared.

This doesn't fully remove hamcrest because it's still used by Mockery, but it will be removed in v3.

Triggered by #3541, which will autoclose after this

#patch

## Approach

- `MatcherAssert::assertThat(..., Matchers::equalTo(...))` can be replaced with `$this->assertEquals`
- `Matchers::equalTo` can be replaced with `$this->equalTo`
  - This is useful for comparing two objects that are different instances of the same class but should have identical values
- `Matchers::anInstanceOf` can be replaced with `$this->isInstanceOf`
- `Matchers::startsWith` can be replaced with `$this->stringStartsWith`
- The iterative approach in FeedbackTest can be replaced with direct assertions

I also added laminas-view to the majors ignorelist due to its dependency on laminas-servicemanager v4.

### DIFF
--- a/.idea/php.xml
+++ b/.idea/php.xml
@@ -219,7 +219,6 @@
       <path value="$PROJECT_DIR$/service-api/vendor/guzzlehttp/guzzle" />
       <path value="$PROJECT_DIR$/service-api/vendor/guzzlehttp/promises" />
       <path value="$PROJECT_DIR$/service-api/vendor/guzzlehttp/psr7" />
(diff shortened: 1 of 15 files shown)


acsauk (approved):
Nice nice nice :shipit:

PR #3556, opened 2026-06-26
LPAL-1754: Account for safelink previsit in ConfirmRegistration

acsauk:
## Purpose

Fixes LPAL-1754

### DIFF
--- a/cypress/e2e/UseMyDetails.feature
+++ b/cypress/e2e/UseMyDetails.feature
@@ -6,7 +6,7 @@ Feature: UseMyDetails
 
   Background:
     Given I ignore application exceptions
-    And I log in as seeded user
+    And I log in as second seeded user
 
   Scenario: Searching for an LPA by donor name returns multiple results (LPAL-539)
     Given I visit the dashboard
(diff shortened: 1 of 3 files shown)


service-front/mezzio/test/AppTest/Handler/ConfirmRegistrationHandlerTest.php
```diff
@@ -42,6 +44,15 @@ private function createRequest(string $token = ''): ServerRequest
             ->withAttribute(SessionMiddleware::SESSION_ATTRIBUTE, $session);
     }
 
+    public function testHeadRequestsReturn200(): void
+    {
+        $request = new ServerRequest([], [], '/signup/confirm/1234', 'HEAD');
+        $response = $this->handler->handle($request);
+
+        self::assertInstanceOf(Response::class, $response);
+        self::assertEquals(StatusCodeInterface::STATUS_OK, $response->getStatusCode());
```
gregtyler, line 53: You might want to add a check that the body content is empty, to make sure this isn't triggering the normal GET request?

acsauk: Good call 👍

gregtyler (dismissed):

cypress/e2e/UseMyDetails.feature
```diff
@@ -6,7 +6,7 @@ Feature: UseMyDetails
 
   Background:
     Given I ignore application exceptions
-    And I log in as seeded user
```
acsauk, line 9: I noticed there is a potential race condition here linked with `UserPersonalDetails.feature` where the address could be updated before this test passes. I've scoped it to the other seeded user instead which, while not ideal, should reduce test re-runs for now.

gregtyler (approved):

PR #3566, opened 2026-07-01
LPAL-2162 refactor service front mezzio to be the root content for service front

allenannom:
LPAL-2162
This PR replaces the legacy laminas-mvc service-front application with the Mezzio app, which was previously running in parallel at service-front/mezzio/. The Mezzio app is now the application at the service-front/ root

the temporary lpa_front_mezzio_app ECR repo is no longer referenced


docker-compose.yml
```diff
@@ -150,7 +157,7 @@ services:
       XDEBUG_CONFIG: client_host=host.docker.internal
       XDEBUG_MODE: ${XDEBUG_MODE:-develop,debug,coverage}
     healthcheck:
-      test: ["CMD", "/usr/local/bin/health-check.sh"]
+      test: ["CMD-SHELL", "cgi-fcgi -bind -connect 127.0.0.1:9000 || exit 1"]
```
gregtyler, line 160: We should continue using `health-check.sh` to emulate the production behaviour

Edit: this can be reinstated on a subsequent PR if that's easier though, given the size of this move. Not a blocker.

andrewpearce-digital (approved):
terraform and actions look good

docker-compose.yml
allenannom: will create a ticket to do this right after if that's okay? unless there's more changes in this PR then I can change it all together

acsauk (approved):
:shipit:

PR #3585, opened 2026-07-06
LPAL-2213 500 unhandled response

allenannom:
LPAL-2213 
getApplication() returns false when the API can't find the LPA (e.g., already deleted, no longer owned by user). The handler was passing false directly to the template, causing Twig to resolve lpa.id as null, which crashed formatLpaId(int $id).

RedisClientFactory created a RedisClient but never called setLogger().

### DIFF
--- a/service-front/src/App/src/View/Twig/LegacyCompatExtension.php
+++ b/service-front/src/App/src/View/Twig/LegacyCompatExtension.php
@@ -139,8 +139,12 @@ class LegacyCompatExtension extends AbstractExtension
         return $number . 'th';
     }
 
-    public function formatLpaId(int $id): string
+    public function formatLpaId(?int $id): string
     {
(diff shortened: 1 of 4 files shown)


service-front/src/App/src/View/Twig/LegacyCompatExtension.php
```diff
@@ -139,8 +139,12 @@ public function ordinalSuffix(int $number): string
         return $number . 'th';
     }
 
-    public function formatLpaId(int $id): string
+    public function formatLpaId(?int $id): string
```
acsauk, line 142: Could we get a test for this now we've added some logic outside of wrapping `ormatter::id($id)`? 🙏

acsauk (dismissed):
:shipit:

acsauk (approved):
:shipit:

PR #3593, opened 2026-07-07
LPAL-1748 - refactor variables to be env centric not account centric

andrewpearce-digital:
## Purpose

To enable us to define different configurations for environments in the development account (for example a demo environment).

Fixes LPAL-1748

## Approach

- refactor 'accounts' variable to be 'environments'
- rename `development` to `default` configuration
- remove account_mapping in favour of naming the account in each environment configuration


acsauk (approved):
:shipit:

PR #3595, opened 2026-07-07
LPAL-2160: Use cognito for admin auth

acsauk:
## Purpose

Fixes LPAL-2160


acsauk:
Things got very messy during a rebase on main - closing in favour of #3625

PR #3596, opened 2026-07-07
LPAL-2200 - Demo Environment

andrewpearce-digital:
## Purpose

Set up a persistent production like environment

Fixes LPAL-2200

## Approach

- add a definition of a demo environment pointed at other demo environments
- deploy demo on the path to live workflow


acsauk (approved):
:shipit:

PR #3605, opened 2026-07-09
LPAL-1960 - Add sin-bin style IP blocking to WAF

andrewpearce-digital:
## Purpose

Block IP addresses that scan for vulnerabilities

Fixes LPAL-1960

## Approach

- add lambda code for ip blocking
- test function on PR and Merge Queue
- lock python dependencies with uv
- only deploy lambda if it passes unit tests
- deploy lambda function
- add waf rules for ip blocking in count mode
- support current region in lambda boto clients

## Learning

- https://github.com/asdf-community/asdf-uv
- https://docs.astral.sh/uv/guides/projects/#uvlock


acsauk (approved):
:shipit:

PR #3613, opened 2026-07-13
LPAL-2247 upgrades to partially fix build, extend snyk ignore

nickdavis2001:
## Purpose

_partially fix snyk finding and extend ignore_

## Approach

_composer upgrade for symfony/validator, but also extend snyk ignore as this is not a complete resolution_

### DIFF
--- a/service-front/composer.json
+++ b/service-front/composer.json
@@ -72,7 +72,8 @@
         "michelf/php-markdown": "^1.9.1",
         "monolog/monolog": "^3.9",
         "php-http/guzzle7-adapter": "^1.0.0",
-        "symfony/validator": "^8.1"
+        "symfony/validator": "^8.1",
+        "symfony/yaml": "^7.4"
     },
     "require-dev": {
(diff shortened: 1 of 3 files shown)


service-front/composer.json
```diff
@@ -72,7 +72,8 @@
         "michelf/php-markdown": "^1.9.1",
         "monolog/monolog": "^3.9",
         "php-http/guzzle7-adapter": "^1.0.0",
-        "symfony/validator": "^8.0"
+        "symfony/validator": "^8.1",
+        "symfony/yaml": "^7.4"
```
acsauk, line 76: Is this an app dependency or a dev dependency? I'm not sure we parse any yaml using that package.

nickdavis2001: Its a depedency of symfony/validator, but upgrading validator on its own didn't fix this, I had to add the extra explicit line for symfony/yaml then this particular issue appeared to resolve

acsauk (approved):
✅

PR #3615, opened 2026-07-13
LPAL-2242 - add a cognito user pool for test users in non-production environments

andrewpearce-digital:
## Purpose

Allow for a test user in non-production environments to test admin features using Cypress

Fixes LPAL-2242

## Approach

- use tfvars to define cognito identity providers for each environment

### DIFF
--- a/terraform/environment/modules/environment/variables.tf
+++ b/terraform/environment/modules/environment/variables.tf
@@ -14,7 +14,9 @@ variable "environment" {
     log_retention_in_days                  = number
     account_name_short                     = string
     associate_alb_with_waf_web_acl_enabled = bool
-    admin_cognito_auth_enabled             = bool
+    cognito = object({
+      admin_cognito_auth_enabled = bool
+    })
     database = object({
       cluster_identifier                 = string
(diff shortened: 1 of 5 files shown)


terraform/environment/modules/environment/variables.tf
```diff
@@ -14,7 +14,9 @@ variable "environment" {
     log_retention_in_days                  = number
     account_name_short                     = string
     associate_alb_with_waf_web_acl_enabled = bool
-    admin_cognito_auth_enabled             = bool
+    cognito = object({
+      admin_cognito_auth_enabled = bool
```
acsauk, line 18: Do we only reference the other object props when we use them?

acsauk (dismissed):
:shipit:

terraform/environment/modules/environment/variables.tf
andrewpearce-digital: yeah, and my intention is to start to group together variables to make them easier to find and use in code

acsauk (approved):
:shipit:

PR #3616, opened 2026-07-13
LPAL-1614: Autowire service-front

acsauk:
## Purpose

Autowire front. Remaining classes generally rely on getting config from the container and setting scalar values or passing the config array in its entirety. We could create some specific config objects that contain the logic of getting the required config from the container so they could be autowired, but that felt out of the scope of this 'quick-win' ticket.

Also moved from setting the logger on classes via `LoggerAwareInterface` to injecting, so we could get rid of simpler factories.

Fixes LPAL-1614


allenannom (dismissed):

Makefile
```diff
@@ -34,44 +34,49 @@ reset:
 
 .PHONY: run-front-composer
 run-front-composer:
-	@docker run --rm -v `pwd`/service-front/:/app/ -v `pwd`/shared/:/shared/ composer:${COMPOSER_VERSION} composer install --prefer-dist --no-interaction --no-scripts --ignore-platform-reqs
+	@docker compose run --rm composer-front install --prefer-dist --no-interaction --no-scripts --ignore-platform-reqs
```
hawx, line 37: extremely minor, but, `front-composer` (etc.) would match the existing naming better

hawx (approved):

Makefile
acsauk: Agree, will bundle with another PR 👍

PR #3618, opened 2026-07-14
LPAL-2160: Add mock Cognito server for local development

acsauk:
## Purpose

Enables LPAL-2160

### DIFF
--- a/mock-cognito/docker/Dockerfile
+++ b/mock-cognito/docker/Dockerfile
@@ -1,0 +1,18 @@
+FROM golang:1.26.4-alpine@sha256:3ad57304ad93bbec8548a0437ad9e06a455660655d9af011d58b993f6f615648 AS build
+
+WORKDIR /app
+COPY go.mod go.sum ./
+RUN go mod download
+COPY . .
+RUN CGO_ENABLED=0 GOOS=linux go build -o /mock-cognito .
+
+RUN CGO_ENABLED=0 GOOS=linux go build -o /healthcheck ./docker/healthcheck.go
+
+FROM scratch AS production
+
+COPY --from=build /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/
+COPY --from=build /mock-cognito /mock-cognito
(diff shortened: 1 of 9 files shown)


mock-cognito/docker/Dockerfile
```diff
@@ -0,0 +1,18 @@
...
+COPY go.mod go.sum ./
+RUN go mod download
+COPY . .
+RUN CGO_ENABLED=0 GOOS=linux go build -o /mock-cognito .
+
+RUN CGO_ENABLED=0 GOOS=linux go build -o /healthcheck ./docker/healthcheck.go
+
+FROM gcr.io/distroless/static AS production
+
+COPY --from=build /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/
```
hawx, line 13: I think you can not do this with the distroless image, or use `scratch` instead

mock-cognito/keys/keygen.go
```diff
@@ -0,0 +1,23 @@
...
+	Public  *rsa.PublicKey
+}
+
+func Generate() (*KeyPair, error) {
+	private, err := rsa.GenerateKey(rand.Reader, 2048)
+	if err != nil {
+		return nil, fmt.Errorf("failed to generate RSA key: %s", err)
+	}
+
+	return &KeyPair{private, &private.PublicKey}, nil
```
hawx: I feel like you could just return the private key, since calling `.PublicKey` isn't that different from `.Public`. And then you could probably just inline this function as it doesn't do much

mock-cognito/docker/Dockerfile
acsauk: I think I originally wanted curl or similar for the healthcheck, then went with the binary. I'll switch to scratch.

mock-cognito/keys/keygen.go
acsauk: Fair, I'll inline to `main()`.

hawx (approved):

PR #3619, opened 2026-07-14
LPAL-2197 no longer commit idea files

nickdavis2001:
## Purpose

_Prevent devs from trampling on each other's settings so that we can configure phpstorm how we want and not randomly lose our settings_

## Approach

_Remove files under .idea from git, and gitignore them. In README mention getting these from a colleague when starting on the project_


acsauk (approved):
:shipit:

PR #3623, opened 2026-07-15
LPAL-2256 fix critical a11y issues,

nickdavis2001:
## Purpose

_The axe tests run within the cypress stitched run were finding critical issues. This PR fixes these so we no longer have a11y violations_

## Issue 1: `aria-allowed-attr` (critical) — 7 pages

**Pages affected:**
- `/lpa/{id}/who-are-you`
- `/lpa/{id}/repeat-application`
- `/lpa/{id}/fee-reduction`
- `/lpa/{id}/how-primary-attorneys-make-decision`
- `/lpa/{id}/when-replacement-attorney-step-in`
- `/lpa/{id}/how-replacement-attorneys-make-decision`
- `/lpa/{id}/applicant`

**Root cause:**
govuk-frontend's Radios component (v6.x) sets `aria-expanded` on radio inputs when they control a conditional reveal section. The `aria-expanded` attribute is not permitted on `role="radio"` per the ARIA spec. This is a known, open bug in govuk-frontend: https://github.com/alphagov/govuk-frontend/issues/979

**Fix:** `service-front/assets/js/opg/govuk-init.js`

Replaced the govuk-frontend `initAll()` call with explicit `createAll()` calls per component. The `Radios` component is replaced by a subclass (`AccessibleRadios`) that overrides `syncConditionalRevealWithInputState()` to omit the invalid `aria-expanded` attribute. The `govuk-radios__conditional--hidden` CSS class (which sets `display:none`) is sufficient to hide content from assistive technologies without any additional ARIA attribute.

> **Note:** Remove the `AccessibleRadios` subclass once govuk-frontend resolves issue #979 upstream.

---

## Issue 2: `label` (critical) — `/lpa/{id}/fee-reduction`

**Root cause:**
In `FeeReductionHandler::buildReductionOptions()`, the first radio option (`reducedFeeReceivesBenefits`) was built with `'id' => 'reductionOptions'` instead of `'id' => 'reducedFeeReceivesBenefits'`. This caused the `<label>` element's `for` attribute to point to the wrong ID, leaving the input without an associated label.

**Fix:** `service-front/src/App/src/Handler/Lpa/FeeReductionHandler.php`

Corrected the id value from `'reductionOptions'` to `'reducedFeeReceivesBenefits'`.

---

## Issue 3: `color-contrast` (serious) — `/lpa/{id}/correspondent`

**Root cause:**
The `.action-edit` button-styled link ("Change correspondent" and similar actions) had a grey background (`#bfc1c3`) with no explicit text colour set. The browser default GOV.UK link colour (`#1d70b8`) against `#bfc1c3` gives a contrast ratio of approximately 2.8:1, below the WCAG AA minimum of 4.5:1.

**Fix:** `service-front/assets/sass/patterns/_person.scss`

Added `color: govuk-colour("black")` (`#0b0c0c`) within the tablet+ media query where the grey background is applied, giving a contrast ratio of ~11:1. Also added `text-decoration: none` (these elements carry `role="button"`) and `text-decoration: underline` on hover.

The `.action-edit` style is project-specific and not part of govuk-frontend; this fix does not deviate from the GOV.UK Design System.

---


acsauk (approved):
Nice job :shipit:

PR #3641, opened 2026-07-17
LPAL-2309 - remove count from dns record now its in the module

andrewpearce-digital:
## Purpose

Updating an existing environment to have the mock onelogin caused an error related to a count on the dns record resource

Fixes LPAL-2309

## Approach

- remove count from dns record (there is a count on the entire module)


allenannom (approved):

acsauk (approved):
:shipit:

PR #3643, opened 2026-07-20
LPAL-2261 Outbound leg of the One Login flow

allenannom:
LPAL-2261
Implements the outbound leg of the One Login flow. When ONELOGIN_ENABLED=true, users are directed to GOV.UK One Login to sign in or create an account instead of the legacy email/password form.


service-api/module/Application/tests/Controller/Version2/Auth/OneLoginControllerTest.php
```diff
@@ -0,0 +1,82 @@
...
+            ->andReturn('')
+            ->once();
+
+        $controller = $this->getOneLoginController();
+        $result     = $controller->startAction();
+
+        $this->assertInstanceOf(ApiProblem::class, $result);
+        $this->assertSame(400, $result->status);
+    }
+
```
acsauk: We probably want to add a test case for the `client_id` as well, and at that point a data provider may be easier to pass the cases in.

service-api/module/Application/src/Controller/Version2/Auth/OneLoginController.php
```diff
@@ -0,0 +1,44 @@
...
+use Application\Model\Service\OneLogin\Service as OneLoginService;
+use Laminas\View\Model\JsonModel;
+use MakeShared\Telemetry\TelemetryEventManager;
+
+/**
+ * @psalm-suppress PropertyNotSetInConstructor
+ */
+class OneLoginController extends AbstractAuthController
+{
+    protected function getService(): OneLoginService
```
hawx, line 15: is there any benefit to having this over using `$this->service` in the 1 place it's needed?

hawx: I see it on the abstract class 🤔, but the typing doesn't actually come through for me without `/** @return OneLoginService */`, so maybe it could be "simplified" to just return the property and type hint the function???

Although since the function isn't used by the abstract class, you could also remove the definition there which would make this typing work (for me, maybe not an issue in PhpStorm)

service-api/module/Application/src/Model/Service/OneLogin/Service.php
```diff
@@ -0,0 +1,115 @@
...
+     */
+    public function __construct()
+    {
+        $this->randomBytes = random_bytes(...);
+    }
+
+    /**
+     * @psalm-suppress PossiblyUnusedMethod
+     */
+    public function setConfig(array $config): void
```
hawx, line 29: thoughts on putting these in the constructor instead, since they are required? (unless this is the Make way, and I need to get used to seeing it)

service-api/module/Application/src/Model/Service/OneLogin/Service.php
```diff
@@ -0,0 +1,115 @@
...
+
+        if (!is_string($discoveryUrl) || $discoveryUrl === '') {
+            throw new RuntimeException('Missing required config: onelogin.discovery_url');
+        }
+
+        $generator = $this->randomBytes;
+
+        $state = rtrim(strtr(base64_encode($generator(12)), '+/', '-_'), '=');
+
+        $nonce = hash('sha256', $generator(24));
```
hawx: I think you could base64 this too? Also probably doesn't have to be so large

hawx: using bin2hex might be another option for these if you want to avoid having to do the URL safe stuff

service-api/module/Application/src/Model/Service/OneLogin/Service.php
```diff
@@ -0,0 +1,115 @@
...
+
+        if ($this->discoveryDocumentFetcher === null && $this->client === null) {
+            throw new RuntimeException('HTTP client must be set via setClient() when no custom DiscoveryDocumentFetcher is provided');
+        }
+
+        $fetcher = $this->discoveryDocumentFetcher
+            ?? new DiscoveryDocumentFetcher($this->client, $discoveryUrl);
```
hawx: I think this class shouldn't have $this->client, and always inject the `DiscoveryDocumentFetcher`

service-front/config/autoload/dependencies.global.php
```diff
@@ -163,6 +163,12 @@ static function (
             Handler\LogoutHandler::class                  => static fn(ContainerInterface $c) => new Handler\LogoutHandler(
                 $c->get('config'),
             ),
+            Handler\LoginHandler::class                   => static fn(ContainerInterface $c) => new Handler\LoginHandler(
+                $c->get(TemplateRendererInterface::class),
+                $c->get(\Laminas\Form\FormElementManager::class),
+                $c->get(\App\Authentication\AuthenticationService::class),
+                getenv('ONELOGIN_ENABLED') === 'true',
```
hawx: I was lazy in my PR, and this can be left too, but maybe next PR we should introduce a `FeatureFlag` class or something to hold this

service-front/src/App/src/Handler/OneLoginSignInHandler.php
```diff
@@ -0,0 +1,41 @@
...
+
+    public function __construct(
+        private readonly OneLoginService $oneLoginService,
+    ) {
+    }
+
+    public function handle(ServerRequestInterface $request): ResponseInterface
+    {
+        $uri         = $request->getUri();
+        $redirectUri = $uri->getScheme() . '://' . $uri->getAuthority() . '/auth/redirect';
```
hawx: if we never want this to work against the real OneLogin on PR envs, then this is fine, otherwise we'll need to be able to set this via an env var, as they don't allow configuring a wildcard for redirect URIs

service-front/src/App/src/Handler/OneLoginSignInHandler.php
```diff
@@ -0,0 +1,41 @@
...
+
+        $result = $this->oneLoginService->start($redirectUri);
+
+        $session = $request->getAttribute(SessionMiddleware::SESSION_ATTRIBUTE);
+        assert($session instanceof SessionInterface);
```
hawx: what do asserts do when they fail? I'd prefer an `if` personally

nickdavis2001 (dismissed):

service-api/module/Application/src/Model/Service/OneLogin/Service.php
allenannom: It's the Make way unfortunately — ServiceAbstractFactory injects these via setter by string key after construction, so they can't be in the constructor without changing the factory pattern across all services.

service-front/config/autoload/dependencies.global.php
allenannom: I'll add something for this

service-api/module/Application/src/Controller/Version2/Auth/OneLoginController.php
allenannom: Will simplify the method. On removing it from the abstract class, getService() is the pattern across all controllers in service-api so I've kept it to stay consistent.

service-front/src/App/src/Feature.php
```diff
@@ -0,0 +1,15 @@
+<?php
+
+declare(strict_types=1);
+
+namespace App;
+
+enum Feature: string
+{
```
hawx, line 8: :+1: neat

hawx (approved):

PR #3644, opened 2026-07-20
LPAL-2327: Prefer PIE to PECL

acsauk:
## Purpose

Fixes LPAL-2327

### DIFF
--- a/service-admin/docker/app/Dockerfile
+++ b/service-admin/docker/app/Dockerfile
@@ -17,6 +17,8 @@ RUN apk add --no-cache icu fcgi \
     && install-php-extensions gmp \
     && rm /usr/bin/install-php-extensions
 
+COPY --from=ghcr.io/php/pie:bin@sha256:65d91cecb73cb1fd03ae2a892d8ec76b25d34468532c0c2032afa9c412cf8b43 /pie /usr/bin/pie
+
 # Install composer dependencies
 COPY --chown=root:root service-admin/composer.json /app/composer.json
 COPY --chown=root:root service-admin/composer.lock /app/composer.lock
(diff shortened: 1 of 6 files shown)


service-pdf/docker/app/Dockerfile
```diff
@@ -14,13 +14,15 @@ RUN apk update \
     && apk add --no-cache openjdk8-jre libpng \
     && apk add --no-cache --virtual .build-dependencies $PHPIZE_DEPS autoconf gcc make musl-dev pkgconfig zlib-dev libpng-dev linux-headers \
     && apk add --upgrade expat \
-    && docker-php-ext-install gd
+    && docker-php-ext-install gd \
+    && docker-php-ext-install sockets
```
acsauk, line 18: I was getting some local test failures related to telemetry, so I needed to install for them to pass (but strange that this doesn't happen in the pipeline 🤔 )

allenannom (dismissed):

service-admin/docker/app/Dockerfile
```diff
@@ -17,6 +17,8 @@ RUN apk add --no-cache icu fcgi \
     && install-php-extensions gmp \
     && rm /usr/bin/install-php-extensions
 
+COPY --from=ghcr.io/php/pie:bin /pie /usr/bin/pie
```
mattmachell: Needs to be sha pinned (same applies to other COPY's of this).

acsauk: Thanks Matt, hadn't realised that was SHA-able. Fixed now.

andrewpearce-digital (approved):

PR #3645, opened 2026-07-20
LPAL-2326 runbook for digital uptake stats

nickdavis2001:
## Purpose

_Make a runbook to enable anyone on the team to be able to update Digital Uptake stats_

## Approach

_runbook as a md file_

### DIFF
--- a/docs/runbooks/updating_digital_uptake_stats.md
+++ b/docs/runbooks/updating_digital_uptake_stats.md
@@ -4,6 +4,8 @@ For the moment this is a manual process.
 
 Clone the repo https://github.com/ministryofjustice/opg-performance-data or pull to update from main if you already have it.
 
+Ensure that pre-commit hooks are installed so that json checking is done. (see README in that repo)
+
 Make a new branch.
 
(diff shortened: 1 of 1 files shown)


allenannom (dismissed):

andrewpearce-digital (changes requested):
remind dev to install pre-commit hooks, they check that the json is good against a schema

hawx:
Is it weird having the runbook on this repo for a different repo? Would it make more sense to have this info on the tickets, then they can just be cloned each time with the different stats?

docs/runbooks/updating_digital_uptake_stats.md
```diff
@@ -0,0 +1,30 @@
+# iUpdating the Digital Uptake stats
```
hawx:
```suggestion
# Updating the Digital Uptake stats
```

nickdavis2001:
> Is it weird having the runbook on this repo for a different repo? Would it make more sense to have this info on the tickets, then they can just be cloned each time with the different stats?

I don't see it does any harm. This was Abi's request. We could put info on the ticket as well though?

andrewpearce-digital (approved):

PR #3649, opened 2026-07-21
LPAL-2257 check for PIE availability before doing a dev (xdebug) build

nickdavis2001:
## Purpose

_Prevent builds breaking due to xdebug having silently failed to install_

## Approach

_Added a pre-flight check in the build_


service-admin/docker/app/Dockerfile
```diff
@@ -17,6 +17,15 @@ RUN apk add --no-cache icu fcgi \
...
     && rm /usr/bin/install-php-extensions
 
+RUN wget -q --timeout=10 -O /dev/null \
+        "https://ghcr.io/token?service=ghcr.io&scope=repository:php/pie:pull" \
+    || { echo ""; \
+         echo "ERROR: The PIE (PHP Installation Engine) registry at ghcr.io is not reachable."; \
+         echo "This build requires the PIE binary from ghcr.io/php/pie to install PHP extensions."; \
+         echo "Check GitHub Container Registry status: https://www.githubstatus.com/"; \
+         echo ""; \
+         exit 1; }
```
acsauk: I'm not sure we need this as I think the build would break anyway on line 29 when we try to copy it if it's not available. The issue we were seeing with PECL was occurring when we tried to pull the xdebug image on line 60.

service-admin/docker/app/Dockerfile
acsauk: I *think* this may do the job:


```suggestion
RUN set -e; if [ "$ENABLE_XDEBUG" = "1" ] ; then \
```

Not tested, but we're explicitly instructing the shell to exit on an error with this flag.

service-admin/docker/app/Dockerfile
nickdavis2001: Take your point about line 29 would fail. So you mean move this block just down to below line 60 to ensure that it hasn't gone AWOL between line 29 running and the attempt to get xdebug?

service-admin/docker/app/Dockerfile
nickdavis2001: OK done :)

service-admin/docker/app/Dockerfile
nickdavis2001: Done :)

acsauk (approved):
:shipit:

PR #3667, opened 2026-07-23
LPAL-2336 - document proces for adding admin users via entra id admin panel

andrewpearce-digital:
## Purpose

Explain how to get access to make's admin service.

Fixes LPAL-2336

## Approach

- Link to guidance in confluence


acsauk (approved):
:shipit:

PR #3670, opened 2026-07-27
LPAL-2313 Handles the return from GOV.UK One Login

allenannom:
Handles the return trip from GOV.UK One Login. Previously the callback was a stub; now a user can sign in through One Login and be routed correctly, with failures handled gracefully instead of erroring.  


- New /v2/auth/onelogin/callback endpoint that exchanges the auth code, validates the ID token, and:                                                                                                                                                                                                                                                                                             
    - Known user (already linked) - issues an auth token and returns their identity.                                                                                                                                                                                                                                                                                                               
    - New user (not yet linked) - returns their One Login sub + email so the front end can offer link-or-create.                                                                                                                                                                                                                                                                                   
  - Email is read from the UserInfo endpoint                                                                                                                                                                                                                                                                        
  - Uses the Facile OIDC client, wired behind a small interface/adapter so it's easy to test.                                                                                                                                                                                                                                                                                                      
                                                                                                                                                                                                                                                                                                                                                                                                   
  **Front end (service-front)**                                                                                                                                                                                                                                                                                                                                                                        
                                                                                                                                                                                                                                                                                                                                                                                                   
  - Callback handler now completes sign-in and redirects to the dashboard (linked) or /link-or-create-account (new user).                                                                                                                                                                                                        
  - Graceful error page ("There is a problem signing you in")                                                                                                                                                                                                            
                                                                                                                                                                                                                                                                                                                                                                                                   
  **Config / local dev**                                                                                                                                                                                                                                                                                                                                                                               
                                                                                                                                                                                                                                                                                                                                                                                                   
  - Wired the One Login signing key via env (ONELOGIN_PRIVATE_KEY, base64-encoded for local so it clears the secret-scanning hooks); real environments inject it from Secrets Manager.                                                                                                                                                                                                             
  - Pinned ONELOGIN_REDIRECT_BASE_URL locally so the redirect URI matches the mock behind the SSL proxy.


service-api/module/Application/src/Model/Service/OneLogin/FacileAuthorizationServiceAdapter.php
```diff
@@ -0,0 +1,40 @@
...
+use Facile\OpenIDClient\Client\ClientInterface;
+use Facile\OpenIDClient\Service\AuthorizationService;
+use Facile\OpenIDClient\Service\UserInfoService;
+use Facile\OpenIDClient\Session\AuthSessionInterface;
+use Facile\OpenIDClient\Token\TokenSetInterface;
+
+final class FacileAuthorizationServiceAdapter implements AuthorizationServiceInterface
+{
+    public function __construct(
+        private readonly AuthorizationService $inner,
```
acsauk: Was there a reason you went with `$inner` over `$authorizationService` or similar?

service-api/module/Application/src/Model/Service/OneLogin/Service.php
```diff
@@ -58,38 +72,151 @@ public function setRandomByteGenerator(callable $generator): void
...
+        $userId = $user->id();
+
+        if ($userId === null) {
+            throw new OneLoginAuthenticationException('missing_user_id');
+        }
+
+        $this->getUserRepository()->updateLastLoginTime($userId);
+
+        if ($user->failedLoginAttempts() > 0) {
+            $this->getUserRepository()->resetFailedLoginCounter($userId);
```
hawx: Would be nice to only do one DB call for these. If you are updating the last login time you can always set failed logins to 0.

service-front/config/autoload/dependencies.global.php
```diff
@@ -163,8 +165,22 @@ static function (
...
+                $c->get(\Laminas\Form\FormElementManager::class),
+                $c->get(\App\Authentication\AuthenticationService::class),
+                App\Feature::OneLogin->isEnabled(),
+            ),
             Handler\OneLoginSignInHandler::class          => static fn(ContainerInterface $c) => new Handler\OneLoginSignInHandler(
-                $c->get(\App\Service\OneLogin\OneLoginService::class),
+                $c->get(OneLoginService::class),
+                $c->get(RedirectUriBuilder::class),
+            ),
+            Handler\OneLoginCallbackHandler::class        => static fn(ContainerInterface $c) => new Handler\OneLoginCallbackHandler(
```
hawx: can this and the `OneLoginSignInHandler` one above be left to autowire?

service-api/module/Application/src/Model/Service/OneLogin/KeyPairManager.php
```diff
@@ -0,0 +1,63 @@
...
+
+declare(strict_types=1);
+
+namespace Application\Model\Service\OneLogin;
+
+use Jose\Component\Core\JWK;
+use Jose\Component\KeyManagement\JWKFactory;
+use RuntimeException;
+
+final class KeyPairManager
```
acsauk: Was there a reason we've made this class final?

service-front/src/App/templates/application/general/auth/index.twig
```diff
@@ -67,6 +67,18 @@
 
 {{ form().closeTag|raw }}
 
+{% if oneLoginEnabled %}
```
hawx: maybe move the "forgot password" link above this, so all the new bits are at the bottom

service-front/config/autoload/dependencies.global.php
```diff
@@ -163,8 +165,22 @@ static function (
...
+                $c->get(TemplateRendererInterface::class),
+                $c->get(OneLoginService::class),
+                $c->get(LoggerInterface::class),
+            ),
```
acsauk: Autowiring should take care of these two classes

service-front/src/App/src/Handler/OneLoginCallbackHandler.php
```diff
@@ -4,24 +4,137 @@
...
 use Psr\Http\Message\ResponseInterface;
 use Psr\Http\Message\ServerRequestInterface;
 use Psr\Http\Server\RequestHandlerInterface;
+use Psr\Log\LoggerInterface;
+use RuntimeException;
 
 class OneLoginCallbackHandler implements RequestHandlerInterface
 {
+    private const string SESSION_KEY_ONELOGIN        = 'onelogin_auth';
+    private const string SESSION_KEY_IDENTITY        = 'identity';
```
acsauk, line 22: Not for this PR, but if this refers to the app identity data rather than something GOL-specific, it feels like we should centralise these consts somewhere/wrap in an object and refer to them rather than scope them per class. As we did the mezzio work it was a bit of a nightmare trying to work out if they all referred to the same thing so we started to collect request attribute consts but never got round to doing anything for session values.

service-front/src/App/src/Service/OneLogin/RedirectUriBuilder.php
```diff
@@ -0,0 +1,24 @@
...
+
+declare(strict_types=1);
+
+namespace App\Service\OneLogin;
+
+use Psr\Http\Message\UriInterface;
+
+final class RedirectUriBuilder
+{
+    public const CALLBACK_PATH = '/auth/redirect';
```
acsauk: Does this need to be public?

service-front/src/App/templates/application/general/auth/onelogin-error.twig
acsauk, line 1: May be worth updating the filename to match the intent of the template here (unless we're planning on it being used for something else other than showing a something went wrong page?)

service-front/src/App/src/Handler/OneLoginCallbackHandler.php
allenannom: yeah that's true - good point. I look to implement this in a future PR

service-front/src/App/src/Service/OneLogin/RedirectUriBuilder.php
allenannom: made it public so the test can access RedirectUriBuilder::CALLBACK_PATH

hawx: I don't think there's much value in a test for a constant value being what it is

acsauk (dismissed):
:shipit:

acsauk (approved):
:shipit:

PR #3672, opened 2026-07-27
LPAL-2346 Provide absolute URL to notify templates

hawx:


service-front/src/App/src/Service/Lpa/CommunicationFactory.php
```diff
@@ -14,13 +14,11 @@ class CommunicationFactory
 {
     public function __invoke(ContainerInterface $container): Communication
```
acsauk: I *think* autowiring should take care of this now its all injected.

acsauk (dismissed):
:shipit:

acsauk (approved):
:shipit:

PR #3681, opened 2026-07-30
LPAL-2231 Show self when managing own shared space

hawx:


service-front/src/App/src/Handler/ManageSharedSpaceHandler.php
allenannom, line 1: Can we add a unit test for this handler?

hawx: I have become too used to codecov pointing out what is missing

acsauk (approved):
:shipit:

PR #3684, opened 2026-07-31
LPAL-2262 — Route One Login users based on whether their account is linked to Make 

allenannom:
LPAL-2262 

included sharedspaceid in onelogin flow

verified the linked identity is established correctly and honours the pre-auth return URL

LinkAccountHandler now uses the verified One Login sub from the session (was a hardcoded placeholder)

missing/expired sign-in session redirects back to /login

added to unit test
added to cypress


service-front/src/App/src/Handler/LinkOrCreateAccountHandler.php
```diff
@@ -41,7 +61,7 @@ public function handle(ServerRequestInterface $request): ResponseInterface
             if ($form->isValid()) {
                 $redirectUrl = $form->get('choice')->getValue() === 'link'
                     ? '/link-account'
-                    : 'TODO-create-account';
+                    : '/signup';
```
hawx: Maybe leave this as the TODO. We want to create them an account instead of asking them to set an email+password (I added https://opgtransform.atlassian.net/browse/LPAL-2360 because I couldn't see an existing ticket)

service-front/src/App/src/Handler/LinkAccountHandler.php
```diff
@@ -48,8 +72,24 @@ public function handle(ServerRequestInterface $request): ResponseInterface
                     ->setPassword($form->get('password')->getValue())
                     ->authenticate();
 
-                if ($result->isValid() && $this->userDetails->setOneLoginSub('TODO-get-the-current-one-login-sub')) {
-                    return new RedirectResponse('/user/dashboard');
+                if ($result->isValid()) {
+                    if ($this->userDetails->setOneLoginSub($sub)) {
```
hawx: One thing I didn't do in this before is check that the account has previously been linked. I've been thinking about removing the password from an account when the sub is set, how do we feel about that?

service-front/src/App/src/Handler/Traits/OneLoginPendingLinkTrait.php
```diff
@@ -0,0 +1,22 @@
+<?php
+
+declare(strict_types=1);
+
+namespace App\Handler\Traits;
+
+use Mezzio\Session\SessionInterface;
+
+trait OneLoginPendingLinkTrait
```
hawx: I don't like traits. But also, I think this might work better as a service that owned the shape of the data, as it is doing a lot of checks for what `$pendingLink` looks like. 

Imagine:

```php
class OneLoginSessionManager
{
  private const string SESSION_KEY = ...;
  public function setPendingLink(string $sub, string $email): void { ... }
  public function getPendingLink(): array { ... }
}
```

hawx: could also make the return type a little DTO, if we wanted

service-front/src/App/templates/application/authenticated/linking/link-or-create-account.twig
```diff
@@ -42,7 +42,7 @@
         </fieldset>
       </div>
 
-      <button class="govuk-button" type="submit">Continue</button>
+        <button class="govuk-button" type="submit" data-cy="link-or-create-continue">Continue</button>
```
hawx, line 45: is there value in `data-cy`, when we could have a step of "Submit the form" that gets the `type=submit` button and clicks it?

service-front/src/App/src/Handler/LinkAccountHandler.php
hawx: also feel free to delay this to another ticket

allenannom: Removing the password from an account when a sub is set is a nice idea, I'm happy to do that. I'll implement it in my current ticket

hawx (approved):

PR #3686, opened 2026-08-03
LPAL-2356: Add user + LPA fixture endpoint behind feature flag

acsauk:
## Purpose

This adds scaffolding for some dynamic fixture creation to allow for re-running Cypress tests in pipelines following state changes that happen in tests that would cause re-runs to fail.

Fixes LPAL-2356

## Approach

I've gone with existing Service/API calls, but we could pivot to making a specialised API fixture handler if we find this is slow.


terraform/environment/modules/environment/ecs_front.tf
```diff
@@ -267,7 +267,8 @@ locals {
         { name = "OPG_LPA_TELEMETRY_REQUESTS_SAMPLED_FRACTION", value = var.environment.telemetry_requests_sampled_fraction },
         { name = "AWS_REGION", value = data.aws_region.current.region },
         { name = "ONELOGIN_ENABLED", value = tostring(var.environment.feature_flags.onelogin_enabled) },
-        { name = "SHARED_SPACES_ENABLED", value = tostring(var.environment.feature_flags.shared_spaces_enabled) }
+        { name = "SHARED_SPACES_ENABLED", value = tostring(var.environment.feature_flags.shared_spaces_enabled) },
+        { name = "CYPRESS_FIXTURES_ENABLED", value = tostring(var.environment.feature_flags.cypress_fixtures_enabled) }
```
andrewpearce-digital: I think it would be good to dpubly make sure this is never enabled on production with a condition.

Something like

name = "CYPRESS_FIXTURES_ENABLED", value =  environment-name != "production" && var.environment.feature_flags.cypress_fixtures_enabled ? tostring(var.environment.feature_flags.cypress_fixtures_enabled) : tostring(false)

cypress/e2e/common/user_with_lpas_fixtures.js
```diff
@@ -0,0 +1,86 @@
...
+    }
+  });
+});
+
+Given(`I create a new user with {int} LPAs`, (lpaCount) => {
+  createUserWithLpas(lpaCount, 'property-and-financial').then(
+    ({ email, password, lpaIds }) => {
+      cy.wrap(email).as('fixtureUserEmail');
+      cy.wrap(password).as('fixtureUserPassword');
+      cy.wrap(lpaIds).as('fixtureUserLpaIds');
```
hawx: `cy.wrap({ email, password, lpaIds }).as('fixtureUser')` thoughts?

service-front/src/App/src/Service/Fixture/CypressFixtureService.php
```diff
@@ -0,0 +1,201 @@
...
+     * @return array{email: string, password: string, userId: string, lpaIds: array<int, string>}
+     */
+    public function createUserWithLpas(int $lpaCount, string $lpaType): array
+    {
+        $email    = $this->generateUniqueEmail();
+        $password = $this->generatePassword();
+
+        $userId = $this->createAndActivateUser($email, $password);
+
+        ['token' => $token] = $this->authenticate($email, $password);
```
hawx, line 42: I had no idea they added this syntax to PHP

service-front/src/App/src/Service/Fixture/CypressFixtureService.php
```diff
@@ -0,0 +1,201 @@
...
+            'email'      => ['address' => 'opglpademo+nancygarrison@example.org'],
+            'canSign'    => false,
+        ]);
+
+        return $lpaId;
+    }
+
+    private function generateUniqueEmail(): string
+    {
+        $identifier = (string) (int) (microtime(true) * 1000.0) . (string) random_int(100000000, 999999999);
```
hawx, line 192: that's a lot of type casting

hawx (dismissed):

service-front/src/App/src/Service/Fixture/CypressFixtureService.php
acsauk: Yeah it was a while back now. I remember being excited that I could hack in [multiple returns](https://www.compilenrun.com/docs/language/php/php-arrays/php-array-destructuring/#example-2-returning-multiple-values-from-a-function) from funcs like Go (but obvs typehinting isn't as helpful)

service-front/src/App/src/Service/Fixture/CypressFixtureService.php
acsauk: Blame Psalm 🙄

cypress/e2e/common/user_with_lpas_fixtures.js
acsauk: So much nicer, bye-bye nested `cy.get()`s 👋

hawx (approved):

radhiashah (approved):

PR #3687, opened 2026-08-03
LPAL-2207 - update prod smoke test script to test internal facing url

andrewpearce-digital:
## Purpose

Allow smoke test to pass when maintenance mode is enabled

Fixes LPAL-2207

## Approach

- use the front fully qualified domain name (fqdn) for smoke test instead of the public facing fqdn


acsauk (approved):
:shipit:

PR #3691, opened 2026-08-03
LPAL-2118 | region changes - Removing access log encryption key and resources

radhiashah:
## Purpose

- Removing resources for access log encryption key - no references found to usage 

Fixes LPAL-2118

## Approach

## Learning


terraform/region/refactor.tf
```diff
@@ -70,3 +70,31 @@ import {
   to = aws_iam_role_policy.vpc_flow_logs
   id = "vpc_flow_logs:vpc_flow_logs"
 }
+
+removed {
```
andrewpearce-digital: these blocks aren't really needed as removing the configuration destroys the resources anyway.

A good use case for the removed block is to remove a resource from state without destroying it

andrewpearce-digital (approved):

PR #3692, opened 2026-08-04
LPAL-2228 Invite member to a shared space

hawx:


service-api/module/Application/src/Controller/Version2/Auth/SharedSpaceController.php
```diff
@@ -132,8 +121,56 @@ public function lpasAction(): Json|ApiProblem
...
+
+        $response = [
+            'members' => $this->sharedSpaceService->getMembers($result['sharedSpaceId']),
+            'invites' => $this->sharedSpaceService->getInvites($result['sharedSpaceId']),
+        ];
+
+        return new Json($response);
+    }
+
+    public function inviteAction(): Json|ApiProblem
```
acsauk, line 202: Should we check that the person inviting is an admin here (the ticket specifically mentions admin status in the first GWT)? I took on board your comment about validating in front only, but it feels more access-control-y?

service-front/src/App/src/Handler/InviteMemberHandler.php
```diff
@@ -0,0 +1,81 @@
...
+    use CommonTemplateVariablesTrait;
+
+    public function __construct(
+        private readonly TemplateRendererInterface $renderer,
+        private readonly FormElementManager $formElementManager,
+        private readonly SharedSpaceService $sharedSpaceService,
+    ) {
+    }
+
+    public function handle(ServerRequestInterface $request): ResponseInterface
```
acsauk, line 34: Similar to API controller, do we want a check for admin status of the person inviting them (and I guess on the template too)?

acsauk (commented):
A question on some access control, but otherwise good to go 👍

service-api/module/Application/src/Controller/Version2/Auth/SharedSpaceController.php
hawx: indeed 🤦, I've copied what you did so the merge will hopefully be ok

acsauk (approved):
:shipit:

PR #3700, opened 2026-08-05
LPAL-2366 - restrict what s3 resources can accessed via the s3 vpc endpoint

andrewpearce-digital:
## Purpose

Apply good policies to VPC endpoints to prevent egress to unauthorised locations

Fixes LPAL-2366

## Approach

- define the s3 resources that can be accessed
- use allowed resources in the S3 gateway endpoint policy

## Learning


acsauk (approved):
:shipit:

PR #3701, opened 2026-08-06
LPAL-2286 - reject linking a Make account already linked to another One Login    

allenannom:
LPAL-2286

associate a correctly authorised make account to one login identity and log them straight into the dashboard
already-linked check which runs before any write
added new handler and generic template for already linked state
added LinkReason.php to be used in service front & api

added tests


service-api/module/Application/src/Model/Service/OneLogin/Service.php
```diff
@@ -241,4 +245,92 @@ public function handleCallback(
...
+        if ($this->sharedSpaceRepository === null) {
+            throw new RuntimeException('SharedSpaceRepository must be set');
+        }
+
+        $user = $this->getUserRepository()->getByUsername($username);
+
+        if (!$user instanceof UserInterface) {
+            $deletionLog = $this->getLogRepository()->getLogByIdentityHash($this->hashIdentity($username));
+
+            return $this->rejectLink(is_array($deletionLog) ? LinkReason::ACCOUNT_DELETED : LinkReason::ACCOUNT_NOT_FOUND);
```
hawx, line 273: we don't really need to display this to the user, so we could simplify and not care?

service-api/module/Application/src/Controller/Version2/Auth/OneLoginController.php
```diff
@@ -92,4 +92,43 @@ public function callbackAction(): JsonModel|ApiProblem
...
+
+        if (!is_string($username) || $username === '') {
+            return new ApiProblem(400, 'username must be provided');
+        }
+        if (!is_string($password) || $password === '') {
+            return new ApiProblem(400, 'password must be provided');
+        }
+        if (!is_string($oneLoginSub) || $oneLoginSub === '') {
+            return new ApiProblem(400, 'oneLoginSub must be provided');
+        }
```
hawx: Like I commented on Alex's last PR, I think we should trust 'front' to pass the correct params since it is just internal

service-api/module/Application/src/Model/Service/OneLogin/Service.php
hawx: the argument for keeping them is that it simplifies investigating things if someone complains, so I don't mind either way

service-api/module/Application/src/Model/Service/OneLogin/Service.php
```diff
@@ -241,4 +245,92 @@ public function handleCallback(
...
+        }
+
+        $this->getUserRepository()->setOneLoginSub($authResult['userId'], $oneLoginSub);
+
+        $this->getLogger()->info('auth.onelogin.link_success', [
+            'user_id' => $authResult['userId'],
+        ]);
+
+        $lastLogin = $authResult['last_login'] instanceof DateTime ? $authResult['last_login'] : new DateTime();
+        $expiresAt = $authResult['expiresAt'] instanceof DateTime ? $authResult['expiresAt'] : new DateTime();
```
hawx: `expiresAt` should always be set

service-api/module/Application/src/Model/Service/Users/Service.php
```diff
@@ -149,6 +149,13 @@ public function createProfile(string $id): ApiProblem|EntityInterface
     public function update(array $data, string $id): ApiProblem|EntityInterface|null
     {
         if (isset($data['oneLoginSub'])) {
+            $existing = $this->getUserRepository()->getById($id);
+            $currentSub = $existing instanceof UserInterface ? $existing->oneLoginSub() : null;
+
+            if (is_string($currentSub) && $currentSub !== '' && $currentSub !== $data['oneLoginSub']) {
```
hawx: is it even possible to get here with a current sub set?

service-front/src/App/src/Service/OneLogin/OneLoginService.php
```diff
@@ -100,4 +100,53 @@ public function callback(
...
+            '/v2/auth/onelogin/link',
+            [
+                'username'    => $email,
+                'password'    => $password,
+                'oneLoginSub' => $oneLoginSub,
+            ],
+            anonymous: true,
+        );
+
+        if (!is_array($result) || !array_key_exists('linked', $result) || !is_bool($result['linked'])) {
```
hawx: similar but opposite argument here, should we just trust 'api' to return what we think it returns?

hawx (dismissed):

service-api/module/Application/src/Model/Service/OneLogin/Service.php
allenannom: I agree, thought it'll be good info to log in case we need to debug

hawx (approved):

PR #3711, opened 2026-08-07
LPAL-2232: Allow admins to suspend users

acsauk:
## Purpose

Fixes LPAL-2232

### DIFF
--- a/cypress/e2e/common/shared_sapce.js
+++ b/cypress/e2e/common/shared_sapce.js
@@ -87,27 +87,44 @@ Given(/^I create a new user with (\d+) LPAs?(?: that belongs to a shared space c
 
 When(`I log in as the newly created fixture user`, () => {
     cy.get('@fixtureUser').then(({email, password}) => {
-        cy.visitWithChecks('/login');
-
-        cy.title().then((title) => {
-            expect(title.toLowerCase()).to.include('sign in');
-        });
-
-        cy.get('[data-cy=login-email]').clear().type(email);
-        cy.get('[data-cy=login-password]').clear().type(password);
-        cy.get('[data-cy=login-submit-button]').click();
+      login(email, password)
     });
+});
+
+When(`I (try to) log in as the member added to the shared space`, () => {
+  cy.contains('Sign Out').click();
+  cy.get('@addedMember').then(({email, password}) => {
+    login(email, password)
+  });
+});
+
+function login(email, password) {
+  cy.visitWithChecks('/login');
+
+  cy.title().then((title) => {
+    expect(title.toLowerCase()).to.include('sign in');
+  });
+
+  cy.get('[data-cy=login-email]').clear().type(email);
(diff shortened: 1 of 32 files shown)


cypress/e2e/common/shared_sapce.js
```diff
@@ -87,40 +87,67 @@ Given(/^I create a new user with (\d+) LPAs?(?: that belongs to a shared space c
...
+  });
+
+  cy.get('[data-cy=login-email]').clear().type(email);
+  cy.get('[data-cy=login-password]').clear().type(password);
+  cy.get('[data-cy=login-submit-button]').click();
+}
+
+Then(`I should not be logged in and I see a suspended account error`, () => {
+  cy.url().should('include', Cypress.config().baseUrl + '/login');
+  cy.contains('This user account has been suspended').should('exist');
```
hawx, line 115: if I liked gherkin/cucumber I might say these should be split, but I don't. I'm now imagining a dreamworld where each feature file is just a `Then <whatever scenario works>` and we stuff `Then('<whatever>')` with all of the Cypress commands required...

service-front/config/autoload/dependencies.global.php
```diff
@@ -168,6 +169,7 @@ static function (
                 $c->get(FormElementManager::class),
                 $c->get(AuthenticationService::class),
                 Feature::OneLogin->isEnabled(),
```
hawx, line 171: I like the neatness of `Feature` being an enum, but am beginning to think it should be a "normal" service class that can be injected and mocked to make the factories go away

service-front/src/App/src/Model/Service/Authentication/Identity/User.php
```diff
@@ -61,6 +61,7 @@ public function __construct(
 
         $this->tokenExpiresIn($expiresIn);
 
+        // TODO can we remove this now admin is handled by Cognito?
```
hawx, line 64: :+1: are you making a ticket to look at this?

hawx (approved):

service-front/src/App/src/Model/Service/Authentication/Identity/User.php
acsauk: I'd forgotten I'd added this todo yesterday - have made https://opgtransform.atlassian.net/browse/LPAL-2371 to look into it.

cypress/e2e/common/shared_sapce.js
acsauk: Could make the feature files less verbose...I dunno if I care enough as I'm already using gherkin in anger 🤷

service-front/config/autoload/dependencies.global.php
acsauk: Will see what that looks like on next PR (unless you are already?)

hawx: I will the next time I have to type it, which may be never tbf

PR #3713, opened 2026-08-10
LPAL-2234: Allow admin users to revoke member invites

acsauk:
## Purpose

Fixes LPAL-2234


hawx (approved):
here comes another set of merge conflicts...

PR #3714, opened 2026-08-11
LPAL-2376: Add more logging to payment flow

acsauk:
## Purpose

Fixes LPAL-2376


allenannom (dismissed):

local-ssl/docker/nginx.conf.template
```diff
@@ -36,11 +36,11 @@ http {
         location / {
             proxy_pass  ${TARGET_SCHEME}://${TARGET_HOST}:${TARGET_PORT};
 
-            proxy_set_header Host              $host;
+            proxy_set_header Host              $http_host;
```
acsauk, line 39: Required so we get ports from localhost when calling `$requestUri->getScheme()` during return URL generation.

hawx (approved):

PR #3720, opened 2026-08-13
LPAL-2360 create new make account for one login when option selected

allenannom:
LPAL-2360

When a user signs in via GOV.UK One Login with no linked Make account and chooses "create a new account", we now mint a passwordless, born-active account and sign them straight in.

New one_login_email column
Link path now also stores one_login_email (setOneLoginSub gained an optional email param)
Contact-email resolution now prefers one_login_email, falling back to the login email.


service-api/module/Application/src/Model/Service/OneLogin/Service.php
```diff
@@ -305,6 +311,59 @@ public function linkExistingAccount(
...
+                'activated'             => $now,
+                'created'               => $now,
+                'last_updated'          => $now,
+                'failed_login_attempts' => 0,
+                'one_login_sub'         => $sub,
+                'one_login_email'       => $oneLoginEmail,
+            ]);
+        } while (!$created);
+
+        $this->getUserRepository()->updateLastLoginTime($userId);
```
hawx, line 344: The only thing this does is set `inactivity_flags => null`, which I'm guessing is the default anyway?

service-front/src/App/src/Handler/LinkOrCreateAccountHandler.php
```diff
@@ -58,22 +65,51 @@ public function handle(ServerRequestInterface $request): ResponseInterface
...
+                    $this->logger->error('auth.onelogin.create_error', ['message' => $e->getMessage()]);
+                    $error = 'api-error';
+                }
             }
         }
 
         return new HtmlResponse($this->renderer->render(
             'application/authenticated/linking/link-or-create-account.twig',
-            array_merge(
-                $this->getTemplateVariables($request),
```
hawx, line 72: I'm guessing this is from the merge of Alex's, use this instead of getting the CSRF token yourself

service-api/module/Application/src/Model/DataAccess/Postgres/UserData.php
```diff
@@ -738,19 +755,17 @@ public function saveProfile(ProfileUserModel $data): void
...
-     *
-     * Once linked, authentication happens in One Login, so the local password is
-     * no longer used — we clear it to avoid keeping a redundant credential.
-     *
      * @param string $userId
      * @param string $oneLoginSub
+     * @param string|null $oneLoginEmail Email supplied by One Login (contact address).
      */
-    public function setOneLoginSub(string $userId, string $oneLoginSub): void
+    public function setOneLoginSub(string $userId, string $oneLoginSub, ?string $oneLoginEmail = null): void
```
hawx: I think you could make email `string`, as I believe you can delete the `if` in `api:Users/Service#update` since `front:UserDetails#setOneLoginSub` isn't used?

service-api/module/Application/src/Model/Service/OneLogin/Service.php
allenannom: It also sets last_login. UserData::create() doesn't have a last_login column in its insert and the column is nullable with no default, so a One Login–created account starts with last_login = NULL. getAccountsInactiveSince() filters on last_login < $since, and that comparison is never true for NULL, so an account created this way and never signed

hawx: ignore me, I clearly didn't read the field names for the dates that were being set

hawx (approved):

PR #3723, opened 2026-08-14
LPAL-2393 - recover payments Gov Pay took but never confirmed back to us  

allenannom:
LPAL-2393

If a user completes payment and never returns to the service (maybe closing the tab as soon as Pay finishes), the redirect that records the payment never happens. The payment is taken but the LPA keeps only a gatewayReference - no payment record, unlocked, and the user can't proceed.

Added payment recovery which now happens on the checkout page itself, so the user only has to open their LPA. Clicking "Continue" from the dashboard routes through /lpa/{id} > lpa/checkout, so both entry paths are covered by one handler.  

Added card payment service which also has re-usable functions
Added Cypress test which uses @CheckoutPaymentGateway and used mock-pay.  Not in dc-up so needs to be re-built 
Added logging for both paths

### DIFF
--- a/service-front/src/App/src/Handler/Lpa/CheckoutPayHandler.php
+++ b/service-front/src/App/src/Handler/Lpa/CheckoutPayHandler.php
@@ -11,19 +11,17 @@ use App\Middleware\RequestAttribute;
 use App\Model\FormFlowChecker;
 use App\Service\Lpa\Application as LpaApplicationService;
 use App\Service\Lpa\Communication;
+use App\Service\Payment\CardPayments;
 use App\Service\Payment\Helper\LpaIdHelper;
 use Fig\Http\Message\RequestMethodInterface;
 use GuzzleHttp\Psr7\Uri;
 use Laminas\Diactoros\Response\RedirectResponse;
 use Laminas\Form\FormElementManager;
-use MakeShared\DataModel\Common\EmailAddress;
 use MakeShared\DataModel\Lpa\Lpa;
-use MakeShared\DataModel\Lpa\Payment\Payment;
 use Mezzio\Helper\UrlHelper;
 use Psr\Http\Message\ResponseInterface;
 use Psr\Http\Message\ServerRequestInterface;
 use Psr\Http\Server\RequestHandlerInterface;
-use Psr\Log\LoggerInterface;
 use RuntimeException;
 
 /**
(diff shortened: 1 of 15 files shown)


service-front/src/App/src/Handler/Lpa/CheckoutPayHandler.php
```diff
@@ -102,24 +100,7 @@ public function handle(ServerRequestInterface $request): ResponseInterface
...
-                $result = $this->lpaApplicationService->updateApplication($lpa->id, ['payment' => $lpa->payment->toArray()]);
-
-                if ($result === false) {
-                    $this->logger->critical('PAYMENT RECORDING FAILED — payment taken but LPA not updated', [
-                        'lpaId'            => $lpa->id,
-                        'gatewayReference' => $gatewayReference,
-                        'has_email'        => $lpa->payment->email !== null,
-                    ]);
-                }
+                $this->cardPayments->recordSuccessfulPayment($lpa, $payment);
```
hawx, line 103: It would slightly change behaviour so no need to do now, but a later refactor could replace all of this with `CardPayments#recoverCompletedPayment` (really it could be `CardPayments#completePayment` since recovery is only due to what it has been introduced to fix)

service-front/src/App/src/Service/Payment/CardPayments.php
```diff
@@ -0,0 +1,151 @@
...
+
+        if ($payment->date !== null || $payment->method !== null) {
+            return false;
+        }
+
+        if ($lpa->locked === true || $lpa->completedAt !== null) {
+            return false;
+        }
+
+        return $lpa->hasFinishedCreation();
```
hawx:
```suggestion
        rreturn $payment instanceof Payment
            && is_string($payment->gatewayReference) && trim($payment->gatewayReference) !== ''
            && $payment->date === null && $payment->method === null
            && $lpa->locked !== true && $lpa->completedAt === null
            && $lpa->hasFinishedCreation();
```

I think it is more readable without so many negatives

service-front/src/App/src/Service/Payment/CardPayments.php
```diff
@@ -0,0 +1,151 @@
...
+                'govpay_status'    => $govPayPayment->state->status ?? 'unknown',
+                'has_email'        => $lpa->payment->email !== null,
+            ]);
+
+            return false;
+        }
+
+        return true;
+    }
+}
```
hawx, line 139: this class is nice, very needed

hawx (dismissed):

hawx (approved):

PR #3724, opened 2026-08-14
Add Telemetry back in to frontend container

gregtyler:
## Purpose

During the move to mezzio we turned off telemetry in the frontend container, which also meant the API container stopped reporting anything. Turn it back on so we can monitor performance.

Fixes LPAL-2395 #patch

## Approach

- Middleware to start and close the root segment
- Dependency configuration to build the Tracer
- Pass the tracer to the API client to propagate context
- Use PSR classes rather than aliases in the ExporterFactory

### DIFF
--- a/service-front/config/autoload/dependencies.global.php
+++ b/service-front/config/autoload/dependencies.global.php
@@ -53,6 +53,8 @@ use Laminas\Stratigility\Middleware\ErrorHandler;
 use MakeShared\Logging\LoggerFactory;
 use MakeShared\Logging\RequestLoggingMiddleware;
 use MakeShared\Logging\RequestLoggingMiddlewareFactory;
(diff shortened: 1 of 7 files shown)


terraform/region/modules/region/modules/vpc_endpoints/variables.tf
```diff
@@ -34,6 +34,12 @@ variable "dynamodb_endpoint_enabled" {
   default     = true
 }
 
+variable "xray_endpoint_enabled" {
```
andrewpearce-digital: Don't need this to be a feature toggle, just enable it, we want it always on

gregtyler: Moved to #3726

andrewpearce-digital (approved):

PR #3730, opened 2026-08-18
LPAL-2289 password reset review

allenannom:
LPAL-2289 
added autocomplete
added the show pwd toggle as it is in other email+password forms
returns 'account-not-activated' so confirmatin page stops lying
added more scenarios to cypress around password reset

reset links now expire properly - `UserData::getByPasswordResetToken` rejects a token past its 24-hour TTL. Previously an expired token still resolved a user. The invalid-token page gained a matching bullet: "not used the link within 24 hours of asking for it".


service-front/src/App/templates/application/general/forgot-password/reset-password.twig
```diff
@@ -40,7 +47,18 @@
     <fieldset>
         <legend class="govuk-visually-hidden">Reset your password</legend>
 
-        {{ macros.formGroup(password) }}
+        <div class="govuk-form-group {{ password.getMessages|length >0 ? 'govuk-form-group--error'}}">
+            {{ formElementErrorsV2(password) }}
```
hawx: errors should go below the label

service-front/src/App/templates/application/general/forgot-password/reset-password.twig
```diff
@@ -40,7 +47,18 @@
     <fieldset>
         <legend class="govuk-visually-hidden">Reset your password</legend>
```
hawx: you could remove the current heading, and make this visible as the heading

hawx: I don't know if this should be a fieldset, especially since it just repeats the title here

service-front/src/App/templates/application/general/forgot-password/reset-password.twig
hawx: a `<button>` instead?

hawx (dismissed):

hawx (approved):

PR #3733, opened 2026-08-19
LPAL-2378 - Boundaried CI role for Terraform in product accounts

andrewpearce-digital:
## Purpose

Use boundaried CI roles for terraform steps

Fixes LPAL-2378

## Approach

- use boundaried ci role in environment
- use boundaried ci role in region
- apply non-ci permission boundary to every iam role


acsauk (dismissed):
:shipit:

acsauk (approved):

PR #3738, opened 2026-08-20
LPAL-2317: Send email when suspending member

acsauk:
## Purpose

Fixes LPAL-2317

### DIFF
--- a/service-api/module/Application/src/Model/DataAccess/Postgres/SharedSpaceData.php
+++ b/service-api/module/Application/src/Model/DataAccess/Postgres/SharedSpaceData.php
@@ -9,6 +9,7 @@ use Application\Model\Service\SharedSpace\MemberNotInSharedSpaceException;
 use Application\Model\Entity\MemberInvite;
 use DateTime;
 use Laminas\Db\Adapter\Exception\InvalidQueryException;
+use Laminas\Db\Sql\Expression;
 use MakeShared\DataModel\SharedSpace\SharedSpaceMember;
 
 class SharedSpaceData extends AbstractBase implements SharedSpaceRepositoryInterface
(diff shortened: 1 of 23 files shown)


service-api/module/Application/src/Model/Service/SharedSpace/SharedSpaceService.php
```diff
@@ -136,10 +131,14 @@ public function getMember(string $sharedSpaceId, string $memberUserId): ?array
         }
 
         return [
-            'id' => $profile->getId(),
-            'name' => $profile->getName(),
-            'email' => $profile->getEmail(),
-            'lastLoginAt' => $profile->getLastLoginAt(),
+            'id' => $member->getId(),
+            'sharedSpaceName' => $member->getSharedSpaceName(),
+            'user' => [
```
acsauk: This felt like a good idea at the time so we could serialize the member into objects but I'm doubting myself now. May be simpler to keep things array based in the front and just return the shared space name instead?

service-api/module/Application/src/Model/DataAccess/Postgres/SharedSpaceData.php
```diff
@@ -107,24 +107,36 @@ public function getSharedSpaceIdForUser(string $userId): ?string
...
-            'isAdmin'       => (bool) $result->current()['isAdmin'],
-            'isActive'      => (bool) $result->current()['isActive'],
-            'createdAt'     => $result->current()['created'],
+            'sharedSpaceName' => $result->current()['name'],
+            'sharedSpaceId'   => $sharedSpaceId,
+            'userId'          => $result->current()['userId'],
+            'isAdmin'         => (bool) $result->current()['isAdmin'],
+            'isActive'        => (bool) $result->current()['isActive'],
+            'createdAt'       => $result->current()['created'],
+            'email'           => $result->current()['identity'],
```
hawx: this will need to take account of the `one_login_email`

hawx (approved):

PR #3741, opened 2026-08-21
LPAL-2406 One Login Flow fixes

allenannom:
LPAL-2406
                                                                                                                                                                                                                                                                                                                                                                                          On the unlinked branch of the callback, the session was regenerated but never cleared. On a shared machine, if user A was signed in and didn't sign out, user B authenticating through One Login arrived at /link-or-create-account with A's identity still active and could reach A's dashboard. 

Added the signed-in guard to /auth/onelogin that LoginHandler and OneLoginHandler already had.

The discovery document and signing keys were cached in an ArrayAdapter, which lives and dies with a single PHP process. Every sign-in made two extra outbound calls to One Login. Now backed by APCu, so it's roughly once per container per hour. Falls back to the old adapter if APCu is unavailable


service-front/src/App/src/Handler/OneLoginSignInHandler.php
```diff
@@ -26,16 +27,20 @@ public function __construct(
...
 
+        if ($session->has(self::SESSION_KEY_IDENTITY)) {
+            return new RedirectResponse('/user/dashboard');
+        }
+
+        $redirectUri = ($this->redirectUriBuilder)($request->getUri());
+
+        $result = $this->oneLoginService->start($redirectUri);
+
         $session->set(self::SESSION_KEY_ONELOGIN, [
```
hawx, line 44: would it be good to clear the session before this is set, so we know it is fresh when they start the process?

hawx: maybe sometime it would be good to refactor the session stuff so it is all handled by a class with distinctly named methods, then we'd be able to trace what is happening easier

hawx (approved):

PR #3750, opened 2026-08-25
LPAL-2426 Inactive-account warnings never reach GOV.UK One Login users

allenannom:
LPAL-2426 ..also removing sub and email when account is deleted


service-api/module/Application/src/Model/Service/AccountCleanup/Service.php
```diff
@@ -142,13 +155,30 @@ private function sendWarningEmails(string $warningType)
...
+                $recipient = $user->username();
+            }
+
             // Tell users the day before, giving them that full day to login.
             $notificationDate = $user->lastLoginAt()->add(\DateInterval::createFromDateString('+9 months'));
 
             try {
                 // Call the notify to send the reminder email - this will thrown an exception on any errors
-                $this->notifyClient->sendEmail($user->username(), $templateId, [
+                $this->notifyClient->sendEmail($recipient, $templateId, [
```
hawx: I think this can safely be simplified to `$user->contactEmail()`, and not do the checks above. It will only return a one login email if they have one set, which will only be if one login is enabled, and once we have a single one login account we'll never be able to turn that envvar off.

allenannom: Simplified it a bit, removed most of the checks but kept the null/invalid-email guard scoped to One Login accounts.

hawx: https://github.com/ministryofjustice/opg-lpa/blob/bf9afb653d1e0913f45af8eedc67d436644a3945/service-api/module/Application/src/Model/Service/OneLogin/Service.php#L207 means it should never be null, and I think if we are going to do validation on the email we should do it when we get it from One Login, not here

hawx (approved):

PR #3751, opened 2026-08-25
LPAL-2412 A user's contact email goes stale if they change it at GOV.UK One Login 

allenannom:
LPAL-2412

### DIFF
--- a/service-api/module/Application/src/Model/Service/OneLogin/Service.php
+++ b/service-api/module/Application/src/Model/Service/OneLogin/Service.php
@@ -204,7 +204,9 @@ class Service extends AbstractService
 
         $email = $userInfo['email'] ?? null;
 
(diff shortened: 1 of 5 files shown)


service-api/module/Application/src/Model/Service/OneLogin/Service.php
```diff
@@ -271,6 +275,8 @@ public function linkExistingAccount(
             throw new RuntimeException('SharedSpaceRepository must be set');
         }
 
+        $oneLoginEmail = trim($oneLoginEmail);
```
hawx: should this and the other `trim`s use `normaliseEmail`? Maybe we are doing that at the wrong layer, just do it once at the start when we get the email from One Login

hawx (approved):

PR #3753, opened 2026-08-26
Move listener code from Module.php to Listener classes

gregtyler:
## Purpose

This separates the various listener behaviours and makes them act more like middleware. This should make the code easier to understand, and easier to migrate in the future if we need to.

Fixes LPAL-2144 #patch

## Approach

Moved listener code into classes, made them extend `AbstractListenerAggregate` so that they self-register.

### DIFF
--- a/service-api/module/Application/src/Library/Listener/ContentNegotiationListener.php
+++ b/service-api/module/Application/src/Library/Listener/ContentNegotiationListener.php
@@ -1,0 +1,79 @@
+<?php
+
+namespace Application\Library\Listener;
+
+use Laminas\EventManager\AbstractListenerAggregate;
+use Laminas\EventManager\EventManagerInterface;
+use Laminas\Mvc\MvcEvent;
+use Laminas\Http\Request as HttpRequest;
+use Laminas\Http\Header\Accept;
+use Laminas\Http\Response;
+use ArrayIterator;
+use Application\Library\ApiProblem\ApiProblem;
+use Application\Library\ApiProblem\ApiProblemResponse;
+use Laminas\Http\Header\HeaderInterface;
+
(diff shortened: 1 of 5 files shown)


service-api/module/Application/src/Library/Listener/ContentNegotiationListener.php
```diff
@@ -0,0 +1,79 @@
...
+use Laminas\Mvc\MvcEvent;
+use Laminas\Http\Request as HttpRequest;
+use Laminas\Http\Header\Accept;
+use Laminas\Http\Response;
+use ArrayIterator;
+use Application\Library\ApiProblem\ApiProblem;
+use Application\Library\ApiProblem\ApiProblemResponse;
+use Laminas\Http\Header\HeaderInterface;
+
+class ContentNegotiationListener extends AbstractListenerAggregate
```
hawx, line 16: isn't it a bit silly having this for an internal API that we fully control 🤔

service-api/module/Application/src/Library/Listener/ErrorListener.php
```diff
@@ -0,0 +1,50 @@
...
+     * Listen for and catch ApiProblemExceptions. Convert them to a standard ApiProblemResponse.
+     */
+    public function onError(MvcEvent $e): ?ApiProblemResponse
+    {
+        $response = null;
+
+        // Marshall an ApiProblem and view model based on the exception
+        $exception = $e->getParam('exception');
+
+        if ($exception instanceof ApiProblemExceptionInterface) {
```
hawx, line 32: we can try to remove this thing one day, there are only 7 places that throw an implementing exception

hawx (approved):

PR #3754, opened 2026-08-26
LPAL-2284 Forgot password flow for shared space works without signing out

hawx:


service-front/src/App/templates/application/partials/nav.twig
```diff
@@ -30,16 +30,14 @@
             Menu
           </button>
           <ul class="govuk-service-navigation__list" id="navigation" data-cy="service-nav-list">
-            {% if nav.name %}
```
hawx, line 33: was there a reason for this I've missed?

acsauk: I think this was here to stop the nav showing before they have entered their personal details on account creation to prevent navigating away and ending up with an empty name.

acsauk: Screenshot 2026-08-27 at 11 51 26

hawx: that makes sense, I'll put "Shared space" in the if too

acsauk (approved):
:shipit:

PR #3756, opened 2026-08-26
LPLA-2318: Shared space admin updates

acsauk:
## Purpose

Fixes LPAL-2318

I updated the local PHPUnit setup to be in line with what we do in CI as I was getting a mismatch of behaviour.


service-admin/src/App/src/Handler/LpasHandler.php
```diff
@@ -16,25 +16,28 @@
...
     {
-        $userId = $request->getAttribute('id');
-        $userEmail = $request->getQueryParams()['email'] ?? '';
+        $userId = $request->getAttribute('userId');
+        $sharedSpaceId = $request->getAttribute('sharedSpaceId');
+        $userEmail = $request->getQueryParams()['email'] ?? null;
+        $sharedSpaceName = $request->getQueryParams()['sharedSpaceName'] ?? null;
 
-        if (!isset($request->getQueryParams()['email'])) {
+        if (empty($userId) && empty($sharedSpaceId)) {
```
hawx: I feel like given the routing this won't be happen? Should it check for the query params though instead, like before?

service-admin/src/App/templates/app/user-lpas.phtml
```diff
@@ -3,7 +3,7 @@
 <?php if (empty($lpas)) { ?>
     <p class="govuk-body">No LPAs found for this user.</p>
 <?php } else { ?>
-    <h1 class="govuk-heading-l">LPAs for <?php if (isset($lpaEmail)) { ?> <?= $lpaEmail; ?> <?php } ?></h1>
+    <h1 class="govuk-heading-l">LPAs for <?= $lpaEmail ?? ($sharedSpaceName ?? 'Unknown'); ?></h1>
```
hawx: this should probably never show "Unknown"

hawx (dismissed):

hawx (approved):

PR #3761, opened 2026-08-28
LPAL-2417 support one login back channel logout notifications

allenannom:
LPAL-2417 
added support for logout when a user signs out of One Login in another service, we sign them out in ours 
 - https://docs.sign-in.service.gov.uk/integrate-with-integration-environment/managing-your-users-sessions/#validate-your-logout-token


service-api/module/Application/src/Model/Service/OneLogin/ThrottledJwksProvider.php
```diff
@@ -0,0 +1,43 @@
...
+
+namespace Application\Model\Service\OneLogin;
+
+use Facile\JoseVerifier\JWK\JwksProviderInterface;
+use Psr\SimpleCache\CacheInterface;
+
+/**
+ * Rate-limits JWKS refetches.
+ */
+final class ThrottledJwksProvider implements JwksProviderInterface
```
hawx: why this and not use `CachedJwksProvider`?

service-front/src/App/src/Handler/OneLoginBackChannelLogoutHandler.php
```diff
@@ -0,0 +1,59 @@
...
+            $accepted = $this->oneLoginService->backChannelLogout($logoutToken);
+        } catch (Throwable $e) {
+            $this->logger->error('auth.onelogin.backchannel_logout_failed', [
+                'message' => $e->getMessage(),
+            ]);
+
+            return $this->response(502);
+        }
+
+        return $this->response($accepted ? 200 : 400);
```
hawx: potentially would be useful to log in the 400 situation here, as hopefully they would always be accepted

hawx (approved):

PR #3764, opened 2026-09-01
LPAL-2418: Persist instructions and preferences at the same time

acsauk:
## Purpose

I initially looked at changing the current behaviour of validating the whole API on every save but this would have involved lots of changes over the codebase so I went with persisting both properties at the same time.

Fixes LPAL-2418


service-front/src/App/src/Handler/Lpa/InstructionsAndPreferencesHandler.php
```diff
@@ -61,46 +62,33 @@ public function handle(ServerRequestInterface $request): ResponseInterface
...
 
+                $metadata = $lpa->getMetadata();
+
                 if (
-                    !isset($lpa->metadata)
-                    || !isset($lpa->metadata['instruction-confirmed'])
-                    || $lpa->metadata['instruction-confirmed'] !== true
+                    count($metadata) === 0
+                    || !isset($metadata['instruction-confirmed'])
+                    || $metadata['instruction-confirmed'] !== true
```
hawx, line 91: this could also be done in the same API call, but that is true of many other things too

service-front/src/App/src/Model/FormFlowChecker.php
```diff
@@ -1492,7 +1492,10 @@ private function lpaHasPeopleToNotify(?int $index = null): bool
      */
     private function isLpaPeopleToNotifySatisfied(): bool
     {
-        return ((array_key_exists(Lpa::PEOPLE_TO_NOTIFY_CONFIRMED, $this->lpa->getMetadata())
+        $metadata = $this->lpa->getMetadata();
+
+        return count($metadata) !== 0
+                && ((array_key_exists(Lpa::PEOPLE_TO_NOTIFY_CONFIRMED, $this->lpa->getMetadata())
```
hawx:
```suggestion
                && ((array_key_exists(Lpa::PEOPLE_TO_NOTIFY_CONFIRMED, $metadata)
```

Makefile
```diff
@@ -306,23 +306,28 @@ dc-unit-tests: dc-front-unit-tests dc-admin-unit-tests dc-api-unit-tests dc-pdf-
 
 .PHONY: dc-front-psalm
 dc-front-psalm:
-	@docker compose -f docker-compose.yml -f docker-compose.local.yml run --rm --no-deps front-app-test vendor/bin/psalm --no-cache --force-jit
+	@docker compose build front-app-test
+	@docker compose -f docker-compose.yml run --rm --no-deps front-app-test vendor/bin/psalm --no-cache --force-jit
```
hawx: I think you should be able to make all of these

```suggestion
	@docker compose -f docker-compose.yml run --build --rm --no-deps front-app-test vendor/bin/psalm --no-cache --force-jit
```

hawx (dismissed):

hawx (approved):

PR #3765, opened 2026-09-02
LPAL-2355 One Log In Set up Integration Environment

allenannom:
LPAL-2355

 Points the demo environment at GOV.UK One Login's Integration environment mock-onelogin. All other environments keep the mock.   

New `onelogin_use_mock` feature flag. Demo is set to false (Integration).

Switched from ES256 to RS256 as One Login Integration advertises only RSA algorithms


terraform/region/secrets.tf
radhiashah, line 1: Remove this whole part from PR and put into a different one - it needs to be merged first to create the secret, then can be called.

service-api/module/Application/src/Model/Service/OneLogin/KeyPairManager.php
```diff
@@ -10,7 +10,10 @@
 
 class KeyPairManager
 {
-    private const ALGORITHM = 'ES256';
+    private const ALGORITHMS = [
+        'RSA' => 'RS256',
+        'EC'  => 'ES256',
```
acsauk: Was there a requirement from GOL to support both here? Or is this a testing speed consideration?

radhiashah (dismissed):

terraform/environment/modules/environment/secrets.tf
```diff
@@ -87,3 +87,11 @@ data "aws_secretsmanager_secret" "elasticache_auth_token" {
 data "aws_secretsmanager_secret" "mock_onelogin_client_id" {
   name = "${var.account_name}/mock-onelogin-client-id"
 }
+
+data "aws_secretsmanager_secret" "onelogin_client_id" {
```
acsauk: Does this need to be a secret?

service-api/module/Application/src/Model/Service/OneLogin/AuthorisationClientManager.php
```diff
@@ -16,6 +16,8 @@ class AuthorisationClientManager
 {
     public const CACHE_TTL = 3600;
 
+    public const ID_TOKEN_SIGNING_ALGORITHM = 'ES256';
```
acsauk: I may have missed something, but is this a mismatch from the test RS256 private key you've added to docker-compose.yml?

service-api/module/Application/src/Model/Service/OneLogin/KeyPairManager.php
allenannom: no requirement at all, no ill get that removed

terraform/environment/modules/environment/secrets.tf
allenannom: Strictly no, main reason I included it was to mirror the mock-onelogin-client-id, and local.onelogin_client_id_secret_arn switches between the two based on the onelogin_use_mock flag. 

@acsauk should I move it to tfvars as a plain value? it'd only be a slight change of implementation with how the flag works

service-api/module/Application/src/Model/Service/OneLogin/AuthorisationClientManager.php
allenannom: They're two different keys                                                                                                                                                                                               
ONELOGIN_PRIVATE_KEY in docker-compose is our key. It signs the private_key_jwt client assertion we send to One Login's /token endpoint. 

ID_TOKEN_SIGNING_ALGORITHM = 'ES256' is their key. It's how we verify the ID token One Login sends back to us.

I'll add a comment on the constant

allenannom: Checked the schema and they do also support RS256 for token signing but only for integration & production, not their mock.

acsauk: OK that makes sense. I wonder if we should update our mock to add RS256 at some point 🤔 would mirror what GDS provide at least. Not for now!

terraform/environment/modules/environment/secrets.tf
acsauk: I think that's fine to be a plain string; tfvars.json and docker-compose.yml feel sensible.

acsauk (approved):
:shipit:

PR #3766, opened 2026-09-02
LPAL-2235 Show who is making changes and prevent conflicting updates (for the first few forms)

hawx:
Imagine this, but going through every single Handler+Controller 😪 

It will hopefully look less ugly when done, as there won't be so many nullable arguments.


service-api/module/Application/src/Model/DataAccess/Postgres/ApplicationData.php
```diff
@@ -216,86 +264,118 @@ public function insert(Lpa $lpa): bool
      *
      * @param Lpa $lpa
      */
-    public function update(Lpa $lpa): void
+    public function update(Lpa $lpa, ?int $ifMatchVersion = null): void
```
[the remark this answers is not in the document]
hawx: although I am trusting us to do `$lpa->setUpdatedBy` 🤔

service-front/src/App/src/Handler/Lpa/PrimaryAttorney/PrimaryAttorneyAddHandler.php
```diff
@@ -125,34 +126,40 @@ public function handle(ServerRequestInterface $request): ResponseInterface
...
+                            throw new RuntimeException(
+                                'API client failed to add a primary attorney for id: ' . $lpa->id
+                            );
+                        }
 
-                    if ($isPopup) {
-                        return new JsonResponse(['success' => true]);
-                    }
+                        $ifMatchVersion = $this->replacementAttorneyCleanup->cleanUp($lpa, $ifMatchVersion + 1);
+                        $ifMatchVersion = $this->applicantService->cleanUp($lpa, $ifMatchVersion);
```
acsauk, line 145: Do we need to be assigning to `$ifMatchVersion` here?

tests/python-api-client/createLpa.py
```diff
@@ -81,8 +81,7 @@
 if args.pn:
     addPersonToNotify(lpaId)
 if args.i:
-    setInstruction(lpaId)
-    setPreference(lpaId)
+    setInstructionPreference(lpaId)
```
acsauk, line 84: Ah missed this, thanks 🙏

acsauk (approved):
That felt like a slog 😫 only how many more to go...!?

:shipit:

service-front/src/App/src/Handler/Lpa/PrimaryAttorney/PrimaryAttorneyAddHandler.php
hawx: probably not, but if this handler ever changed it would be easy to miss it so figured may as well 🤷

PR #3768, opened 2026-09-02
LPAL-2344: Vary shared space manage template by admin role

acsauk:
## Purpose

Fixes LPAL-2344

### DIFF
--- a/service-front/src/App/templates/application/authenticated/shared-space/manage.twig
+++ b/service-front/src/App/templates/application/authenticated/shared-space/manage.twig
@@ -1,7 +1,12 @@
 {% import "application/macros.twig" as macros %}
 {% extends 'layout/layout.twig' %}
 
-{%- set pageTitle = 'Manage your Shared Space' -%}
+{% if signedInUserIsAdmin %}
+    {%- set pageTitle = 'Manage Shared Space' -%}
+{% else %}
+    {% set pageTitle = 'View Shared Space' %}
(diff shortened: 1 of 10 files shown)


service-front/src/App/templates/application/authenticated/shared-space/manage.twig
```diff
@@ -59,15 +64,16 @@
...
+      <h1 class="govuk-heading-xl">{{ sharedSpaceName }}</h1>
 
       <div class="govuk-tabs app-tabs--no-border" data-module="govuk-tabs">
         <h2 class="govuk-tabs__title">Contents</h2>
 
         <ul class="govuk-tabs__list" role="tablist">
           <li class="govuk-tabs__list-item govuk-tabs__list-item--selected" role="presentation">
             <a class="govuk-tabs__tab" href="#members" id="tab_members" role="tab" aria-controls="members" aria-selected="true" tabindex="0">
-              Manage members
+              {% if signedInUserIsAdmin %}Manage members{% else %}View members{% endif %}
```
hawx: On the design in figma there are no tabs for the non-admin, as they shouldn't be able to import

acsauk: Ah I'd missed the decision to lock import to admins only. I'd assumed there was a further design ticket to be worked. I'll update the template.

hawx (approved):

PR #3773, opened 2026-09-03
LPAL-2355 correctly handle json from AWS secret

allenannom:
LPAL-2355


radhiashah (approved):

andrewpearce-digital (changes requested):
don't use json here, save the secret as plaintext

PR #3775, opened 2026-09-03
LPAL-2365: Update invite designs and add validation errors

acsauk:
## Purpose

Fixes LPAL-2365


cypress/regressions/baseline/add-attorney-form.png
acsauk, line 1: Side effect of applying `govuk-list` to the error summary to get the expected spacing between errors.

hawx: this is the world's hardest spot the difference

service-api/module/Application/src/Controller/Version2/Auth/SharedSpaceController.php
```diff
@@ -193,7 +194,11 @@ public function inviteAction(): Json|ApiProblem
                 $created->add(DateInterval::createFromDateString('7 days')),
             ));
         } catch (Throwable $e) {
-            return new ApiProblem(StatusCodeInterface::STATUS_INTERNAL_SERVER_ERROR, 'Unable to process request ' . $e->getMessage());
+            return match (true) {
```
hawx: I'm not sure I'm a fan of using `catch`-all + `match` here, when you could `catch` each exception separately

service-api/module/Application/src/Model/Entity/MemberInvite.php
```diff
@@ -35,4 +35,9 @@ public static function create(
     ): MemberInvite {
         return new MemberInvite(null, $userId, $sharedSpaceId, $firstNames, $lastName, $email, $isAdmin, $code, $created, $expires);
     }
+
+    public function getEmail(): string
```
hawx: `email` is already a public readonly property, so don't really need a getter

service-api/module/Application/src/Model/Service/SharedSpace/SharedSpaceService.php
```diff
@@ -293,6 +293,20 @@ public function getInvites(string $sharedSpaceId): array
...
     public function invite(MemberInvite $memberInvite): array
     {
+        $existingInvites = $this->sharedSpaceRepository->getInvites($memberInvite->sharedSpaceId);
+        foreach ($existingInvites as $invite) {
+            if ($invite->getEmail() === $memberInvite->getEmail()) {
+                throw new InviteAlreadyExistsException();
+            }
+        }
+
+        $members = $this->sharedSpaceRepository->getMembers($memberInvite->sharedSpaceId);
```
hawx: we should be doing a query for this, e.g. `->hasMemberWithEmail($spaceId, $email)`, as the shared space might have a lot of members

hawx: similarly above

acsauk: Yeah, I guess I was just being lazy 🤷 will update

service-api/module/Application/src/Model/Entity/MemberInvite.php
acsauk: My OCD twitches at the mix of getters and direct access, but I agree it's not needed.

cypress/regressions/baseline/add-attorney-form.png
acsauk: Glad to provide some light entertainment 🫡 (top tip, the swipe version of the comparison tool is very handy for these tiny changes)

hawx (approved):

PR #3778, opened 2026-09-07
LPAL-2408 ur environment only allow research participants to access the environment

andrewpearce-digital:
## Purpose

Require a login to access the ur environment (like basic auth)

Fixes LPAL-2408

## Approach

- define a ur environment
- add cognito client for ur environment front loadbalancer access
- require oidc authentication on front loadbalancer in the ur environment
- allow ur to be publicly accessible

When a user requests the ur environment, they will be prompted for a username and password. This will allow them to reach the home page.

Sessions will last an hour

Double checked and this doesn't impact production's availability


acsauk (approved):
It may be worth checking UR session length to see if we need to extend past an hour

:shipit:

PR #3779, opened 2026-09-08
LPAL-2235 Check version of LPA when updating to prevent conflicts

hawx:
I've split out the annoying work to a new ticket (https://opgtransform.atlassian.net/browse/LPAL-2493), so just left TODOs where that will impact so they don't get missed.


service-front/src/App/src/Handler/Lpa/ReplacementAttorneyAddTrustHandler.php
```diff
@@ -122,7 +124,7 @@ public function handle(ServerRequestInterface $request): ResponseInterface
                     }
 
                     $html = $this->renderer->render(
-                        'application/authenticated/lpa/replacement-attorney/trust-form.twig',
+                        'Application/authenticated/lpa/replacement-attorney/trust-form.twig',
```
acsauk: Was this meant to be capitalised?

acsauk (dismissed):

service-front/src/App/src/Handler/Lpa/ReplacementAttorneyAddTrustHandler.php
hawx: zoom

good spot in this big PR

acsauk (approved):
:shipit:

PR #3791, opened 2026-09-11
make sure ur is protected by the cleanup script and the workspaces db

andrewpearce-digital:
## Purpose

Don't destroy the ur environment

## Approach

- update cleanup scripts to protect ur environment


acsauk (approved):
:shipit:

PR #3794, opened 2026-09-15
Fix Mock One Login on PR and UR environments

andrewpearce-digital:
## Purpose

Clicking "Sign in with Onelogin" caused the page to hang. Once that was fixed, accessing the mock presented the user with a "no valid certificate" error.

## Approach

- add a security group rule to allow api ecs service to talk to mock onelogin service (mock onelogin was already configured to accept incoming requests)
- add a certificate for mock one login load balancer (will be attached in a subsequent PR)


acsauk (dismissed):
:shipit:

hawx (approved):

PR #3797, opened 2026-09-15
LPAL-2348: Update your details design and change all actor title fields to text input

acsauk:
## Purpose

I ended up updating all title fields given how `Name` and `LongName` validation logic is shared between all forms.   

Fixes LPAL-2348


service-front/assets/js/moj/moj.modules/moj.title-switch.js
acsauk, line 1: 🔪

cypress/e2e/Password.feature
```diff
@@ -7,10 +7,7 @@ Feature: Password
     Scenario: Change password with mismatched details
       Given I ignore application exceptions
       When I log in as seeded user
-      And I visit link containing "Your details"
-      Then I am taken to "/user/about-you"
-      When I visit link containing "Change Password"
-      Then I am taken to "/user/change-password"
+      And I visit "/user/change-password"
```
hawx, line 10: this is quicker, but does mean we aren't showing that navigation works. Personally would prefer to keep one the old way

cypress/e2e/SignupAndChangeDetails.feature
```diff
@@ -70,14 +70,14 @@ Feature: SignupAndChangeDetails
...
-        When I click element marked "Change Password"
+        When I visit "/user/change-password"
         Then I see "Change your password" in the page text
         And I see "Cancel" in the page text
 
         When I visit link containing "Cancel"
         Then I am taken to "/user/about-you"
 
-        When I click element marked "Change Email Address"
+        When I visit "/user/change-email-address"
```
hawx, line 80: I'm also not sure why you've changed these

service-front/assets/js/lpa/templates/postcodeLookup.search-field.html
```diff
@@ -1,4 +1,4 @@
-<div class="form-group js-PostcodeLookup__search">
+<div class="form-group js-PostcodeLookup__search govuk-!-margin-bottom-4">
```
hawx: you could use `govuk-form-group` instead of adding the margin yourself

service-front/assets/js/lpa/templates/postcodeLookup.search-result.html
```diff
@@ -1,4 +1,4 @@
-<div class="form-group">
+<div class="form-group govuk-!-margin-bottom-4">
```
hawx: and here too

service-front/src/App/templates/application/authenticated/about-you/index.twig
```diff
@@ -10,183 +10,200 @@
...
 
+    {% if (oneLoginEnabled) %}
+      <details class="govuk-details">
+        <summary class="govuk-details__summary">
+      <span class="govuk-details__summary-text">
+        How can I change my email address or password?
+      </span>
+        </summary>
+        <div class="govuk-details__text">
+          Your email address and password is managed via GOV one login. If you want to make changes to either then go to your GOV one login.
```
hawx: I know it is in the figma design, but "GOV.UK One Login" is the name of the service

service-front/src/App/templates/application/authenticated/about-you/index.twig
```diff
@@ -10,183 +10,200 @@
...
-        {{ macros.formGroup(title) }}
-        {{ macros.formGroup(firstName) }}
-        {{ macros.formGroup(lastName) }}
+      {{ macros.formGroup(title) }}
+      {{ macros.formGroup(firstName) }}
+      {{ macros.formGroup(lastName) }}
 
     </fieldset>
 
+    {% if (oneLoginEnabled) %}
```
hawx: this text should be using the Shared Space feature flag

shared/module/MakeShared/src/DataModel/Common/LongName.php
```diff
@@ -43,7 +42,6 @@ public static function loadValidatorMetadata(ClassMetadata $metadata)
             new Assert\NotIdenticalTo(''),  // Not identical to en empty string
```
hawx: does this need removing too?

hawx (commented):
maybe for another time, but the fields should probably be labelled "Title (optional)" now

cypress/e2e/Password.feature
acsauk: This was more so not wanting to add another workaround with a query param for the new design. The links won't exist once we go live with GOL, so this felt like a compromise of keeping the feature itself tested but not the link to get there.

service-front/src/App/templates/application/authenticated/about-you/index.twig
acsauk: Fair - I was taking the view that we'll be releasing both at the same time but it makes more sense for this to be scoped to shared spaces.

shared/module/MakeShared/src/DataModel/Common/LongName.php
acsauk: Yes 🙏

service-front/assets/js/lpa/templates/postcodeLookup.search-field.html
acsauk: My eyes saw `form-group` not clocking the `govuk` was missing so assumed there was some weird custom CSS going on.

cypress/e2e/Password.feature
hawx: could have a `@RequiresNotOneLogin` or something instead? But yeah, it won't exist for long and we are unlikely to break them (I hope), so can be ignored

hawx (dismissed):

hawx (approved):

PR #3799, opened 2026-09-16
LPAL-2413 Setting User Agent for One login

allenannom:
LPAL-2413

Explicitly setting User Agent for One Login requests instead of relying on the default - "GuzzleHttp/7"


service-api/module/Application/tests/Model/Service/OneLogin/UserAgentTest.php
```diff
@@ -0,0 +1,160 @@
...
+            $this->assertSame(
+                self::CONFIGURED_USER_AGENT,
+                $transaction['request']->getHeaderLine('User-Agent'),
+            );
+        }
+    }
+
+    public function testShippedConfigMatchesTheDocumentedFormat(): void
+    {
+        $config = include __DIR__ . '/../../../../../../config/autoload/global.php';
```
hawx, line 93: that's a lot of dots

service-api/module/Application/tests/Model/Service/OneLogin/UserAgentTest.php
```diff
@@ -0,0 +1,160 @@
...
+use Laminas\ServiceManager\ServiceManager;
+use PHPUnit\Framework\TestCase;
+
+/**
+ * GOV.UK One Login rejects direct calls that arrive without a User-Agent with a 403, and asks for
+ * the same value on every one of them.
+ *
+ * @see https://docs.sign-in.service.gov.uk/before-integrating/set-user-agent-header/
+ */
+class UserAgentTest extends TestCase
```
hawx, line 25: I'm not sure that this test is necessary, it is basically doing a test of Guzzle configuration working as expected. If we were wrapping with our own class, or configuring a central instance that we retrieved then maybe it would be fine

hawx (approved):

service-api/module/Application/tests/Model/Service/OneLogin/UserAgentTest.php
hawx: I would say we _should_ have a central instance that we `$container->get`, but I've just seen that there are only 2 uses of `new GuzzleClient` in api and then we have 4 in front... ah, how I hate the front/api split

PR #3802, opened 2026-09-16
LPAL-1755| Setting workspace manager duration to 12h instead of default 24h

radhiashah:
## Purpose

Reduce the terraform workspace environment duration
Currently set to 24 hours and could be between 6 and 12 hours

Fixes LPAL-1755

## Approach

Reduce time PR envs are kept online overnight  - cost effective 

Default value is set to 24h - changed in workflow by adding input value and setting to 12h. 

## Learning


.github/workflows/workflow_pr.yml
```diff
@@ -148,6 +148,7 @@ jobs:
       is_ephemeral: false
       workspace_manager_aws_account_id: "050256574573"
       workspace_manager_aws_iam_role: opg-lpa-ci-boundary
+      workspace_manager_protection_duration: "12"
```
andrewpearce-digital: not needed for region, it's not an ephemeral environment

andrewpearce-digital (approved):

PR #3804, opened 2026-09-17
LPAL-2545 | Security hub - adding TLS requests to S3 bucket policy for redacted logs

radhiashah:
S3 general purpose buckets should require requests to use TLS

Fixes LPAL-2545

## Approach

This is enabled in all other S3 buckets in repo except redacted logs bucket. 

## Learning


radhiashah:
S3 bucket will be deleted in LPAL-2368 - not required anymore

PR #3805, opened 2026-09-17
LPAL-2493 Use form post for deleting actors from an LPA

hawx:


service-front/src/App/src/Handler/Lpa/CertificateProvider/CertificateProviderDeleteHandler.php
```diff
@@ -17,6 +17,7 @@
 use Psr\Log\LoggerInterface;
 use RuntimeException;
 
+// TODO(LPAL-2493): Delete this handler after new template has been deployed.
```
acsauk, line 20: Should this (and other similar comments) point to a different ticket reference or was this a reminder during dev?

acsauk (approved):

service-front/src/App/src/Handler/Lpa/CertificateProvider/CertificateProviderDeleteHandler.php
hawx: I'm going to do another PR under the same ticket, it was just a note so it isn't forgotten

